PR #32, opened 2024-08-20
[1] - added tests for winners_and_losers

masterismail:


anth-volk (commented):
Thanks for this, @masterismail. Most of this looks good, but I want to confirm one thing. For the winners and losers tests, it looks like you code a particular tax change, then check the decile output against a hard-coded series of values.

If this is correct, this would make the tests extremely fragile. Because of the complex interaction effects tax policy creates, it wouldn't be surprising if these tests failed, say, a few months down the road because the UK changed how it calculated some particular benefit program. If that's the case, I'd ask that you rewrite these, but I'm not quite sure how yet - I'll leave that in a separate comment.

anth-volk:
What I would instead do, if possible, is inject particular values that you expect to return a particular result. For example, for `ByIncomeDecile`, I would actually define a baseline and reform array (assuming these take arrays) and then either create a tax change for test purposes that just assesses some sort of flat tax or predictably modifies one thing, but isn't present within the overall UK tax & benefit system. Then, I'd test that data and ensure that it returns the expected change.

anth-volk:
Changing the status of this to `change requested`

anth-volk (changes requested):
See comments above

MaxGhenis:
We also don't want the test suite to require restricted data. Would this run over a synthetic file? Or how about creating a simple fake country system for these tests? @nikhilwoodruff

masterismail:
Absolutely, @anth-volk. You’re right—I had a similar conversation with Nikhil about this. We also recognized the fragility of these tests and mentioned it during our meeting. The current tests were really just a temporary solution to ensure that the overall package workflow was functioning correctly for the metrics we defined.

given that we also need to consider that the test suite don't have restricted data, creating a specific country system for testing is a good idea. It could also benefit other packages. Thanks so much for bringing this up!

policyengine/tests/economic_impact/winners_and_losers/test_by_wealth_decile/test_by_wealth_decile.py
```diff
@@ -0,0 +1,40 @@
+import pytest
+import yaml
+import os
+from policyengine import EconomicImpact
+
+def assert_dict_approx_equal(actual, expected, tolerance=1e-4):
```
nikhilwoodruff, line 6: Can we just bump this up to 5e-2? Seems reasonable to me.

policyengine/tests/economic_impact/winners_and_losers/test_by_income_decile/by_income_decile.yaml
```diff
@@ -0,0 +1,19 @@
+# Winners and Losers by income
```
nikhilwoodruff, line 1: Can we change this to a US test?

nikhilwoodruff (changes requested):

policyengine/tests/economic_impact/winners_and_losers/test_by_wealth_decile/test_by_wealth_decile.py
masterismail: done.

policyengine/tests/economic_impact/winners_and_losers/test_by_income_decile/by_income_decile.yaml
masterismail: Is there any specific parameter you would recommend that could impact our target metrics ?

nikhilwoodruff: This one- see `reproduce in python` for the parameter name. https://policyengine.org/us/policy?reform=51488&focus=gov.irs.credits.ctc.refundable.fully_refundable&region=us&timePeriod=2024&baseline=2

masterismail: updated with the requested parameter, although we still have zeroes at some places.

PR #58, opened 2024-12-09
Refactor to improve (but not entirely fix) type hints

nikhilwoodruff:
The existing type hints in simulation.py are incorrect. This attempts to fix them (And associated issues found while doing that), but I had to use # type:ignore in one place related to where we are trying to work on microsimulation and simulation in the same code even though only some of it applies to microsimulations.

Do not merge as-is, this commit is at-request to show current progress

Above is from @mikesmit- I'm opening it as a draft PR to more easily review. Thanks Mike!


nikhilwoodruff:
Superseded by #62

PR #71, opened 2025-01-08
Fix bug in subsetting

### LINKED ISSUE #70: Subsetted simulations don't respect `Simulation.time_period`
After setting e.g. the region to New York, simulations don't reapply the time period setting, causing errors in some cases.

nikhilwoodruff:
Fixes #70


nikhilwoodruff:
I'm going to merge this due to its small size and time-sensitivity.

PR #100, opened 2025-02-14
Add child poverty impact extension

### LINKED ISSUE #99: Add multiple child extension

PavelMakarchuk:
Fixes #99


nikhilwoodruff:
This is super helpful! Notes from our discussion:

* Overall the syntax works well.
* To enable more complex analyses to work well, we should drop the simulation parameters from the `Simulation` initialisation call to the actual function we ask users to define, and instead surface the simulation initialisation functions in `Simulation` to functions.
* We should provide guidance on avoiding overloading the memory limits in the documentation.

PR #116, opened 2025-05-06
Create parameter serializer and convert Infinity and -Infinity to np.inf and -np.inf

### LINKED ISSUE #115: Create reform serializer and deserializer
In policies that set a parameter to infinity, we have at least one simulation workflow that crashes due to an inability to set a parameter to `Infinity`. At this point in execution, this should have already been deserialized from JS into Python as `np.inf` (and `-np.inf` for `-Infinity`). My guess is that this API doesn't use the code present in API v1 for safe JSON deserialization.

We should create a serializer and deserializer that converts reform dicts to `-core`-valid reform dicts. This should include:
* Ensuring that `Infinity` is properly decoded to `np.inf` and `-Infinity` to `-np.inf`
* We may want to convert ints to floats, as we do elsewhere

This will make it easy to add custom rules down the road, as well.

anth-volk:
Fixes #115 

This PR:

1. Converts `"Infinity"` to `np.inf` and `"-Infinity"` to `-np.inf`
2. Adds testing
3. Creates subschemas for the `ParametricReform` schema to explicitly support and validate three reform input types:
```
{
  "param": 15_000
}
```
```
{
  "param": {
    "2024": 15_000
  }
}
```
```
{
  "param": {
    "2024-01-01.2025-12-31": 15_000
  }
}
```

### DIFF
--- a/policyengine/simulation.py
+++ b/policyengine/simulation.py
@@ -195,7 +195,6 @@ class Simulation:
         }[country][macro]
 
         if isinstance(reform, ParametricReform):
-            print("Dumping reform model")
             reform = reform.model_dump()
 
         simulation: CountrySimulation = _simulation_type(
--- a/policyengine/simulation.py
+++ b/policyengine/simulation.py
@@ -195,6 +195,7 @@ class Simulation:
         }[country][macro]
 
         if isinstance(reform, ParametricReform):
+            print("Dumping reform model")
             reform = reform.model_dump()
 
         simulation: CountrySimulation = _simulation_type(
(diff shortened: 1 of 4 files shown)


policyengine/simulation.py
```diff
@@ -195,6 +195,7 @@ def _initialise_simulation(
         }[country][macro]
 
         if isinstance(reform, ParametricReform):
+            print("Dumping reform model")
```
nikhilwoodruff, line 198: Could you remove this?

nikhilwoodruff (approved):
This looks good and yes conforms to what I'd expect the schema to be.

policyengine/simulation.py
anth-volk: Removed

anth-volk:
Will merge after tests pass, given addressing of all issues and pre-emptive approval from Nikhil

PR #126, opened 2025-05-13
Add country package version to economy comparisons

### LINKED ISSUE #125: Add country package version to economy output runs
This is necessary to complete https://github.com/PolicyEngine/issues/issues/366. Add the current country package version to an API simulation run.

anth-volk:
Fixes #125.

## What Changed
- Added a utility function to fetch the version of a given package in an environment
- Added a function to fetch the version of a policyengine country package based upon country ID
- Updated `EconomyComparison` schema to include `country_package_version` key
- Updated `calculate_economy_comparison()` to provide the country package version key as part of its returned `EconomyComparison` dict
- Added tests for utility functions

## Why This Change
- Allows for passing of country package version value to workflow run output, simplifying downstream comparison of this and v1 workflow country package version


tests/utils/test_versioning.py
```diff
@@ -0,0 +1,63 @@
...
+class TestGetCountryPackageVersion:
+    def test__given_country_package_exists__then_return_version(
+        self, patch_importlib_version
+    ):
+        test_country = "us"
+
+        test_version = get_country_package_version(test_country)
+
+        # Version number defined by mock
+        assert test_version == "1.0.0"
```
mikesmit, line 42: **comment, non- blocking** I don't love this magic value of "1.0.0". I have to go look at the fixture to see it's hard coded.

What about something like:
```
assert test_version == patch_importlib_verion.mocked_version
```

**comment, non-blocking** I also don't love 1.0.0 as a value. I'd kind of prefer "MOCK_VERSION_STRING" or something so when an assert fails and the value gets printed out it's really clear this is a mock value.

policyengine/outputs/macro/comparison/calculate_economy_comparison.py
```diff
@@ -10,6 +10,7 @@
 from policyengine.outputs.macro.single.calculate_single_economy import (
```
mikesmit, line 10: **comment, maybe blocking** I don't see a unit test update below for this file? Is that intentional

mikesmit (commented):
One opinionated comment, otherwise nitts or questions.

policyengine/constants.py
```diff
@@ -3,6 +3,29 @@
...
+    "il",
+    "ng",
+]
+
+SUPPORTED_COUNTRY_PACKAGES = [
+    f"policyengine_{country}" for country in SUPPORTED_COUNTRY_IDS
+]
+UNSUPPORTED_COUNTRY_PACKAGES = [
+    f"policyengine_{country}" if country != "ca" else "policyengine_canada"
+    for country in UNSUPPORTED_COUNTRY_IDS
```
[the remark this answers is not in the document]
anth-volk: Alas, the linter broke this up, but fair point, will address

policyengine/outputs/macro/comparison/calculate_economy_comparison.py
anth-volk: This is because the test just doesn't exist. I will add testing to cover relevant portions of my own changes, but am hesitant to bite off the entire method here.

anth-volk:
Update: tests fail due to a typo in the schema definition

policyengine/utils/versioning.py
```diff
@@ -0,0 +1,30 @@
...
+def get_country_package_version(country_id: str) -> str:
+
+    if country_id in UNSUPPORTED_COUNTRY_IDS:
+        raise ValueError(f"Country ID {country_id} is not supported.")
+
+    if country_id not in SUPPORTED_COUNTRY_IDS:
+        raise ValueError(f"Country ID {country_id} is not recognized.")
+
+    try:
+        return version(f"policyengine_{country_id}")
```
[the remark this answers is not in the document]
anth-volk: Fair point. I promise, this wasn't even AI code.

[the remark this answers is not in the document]
anth-volk: Completed

policyengine/constants.py
```diff
@@ -3,6 +3,38 @@
...
+    f"policyengine_{country}" for country in SUPPORTED_COUNTRY_IDS
+]
+
+
+def _package_name_for(country_id: str) -> str:
+    return (
+        f"policyengine_{country_id}"
+        if country_id != "ca"
+        else "policyengine_canada"
+    )
```
mikesmit, line 27: **question** This logic is duplicated below (I.e. manually checking Canada). As we add more countries, etc. someone will have to update both. Can we do this once and re-use?

mikesmit (commented):
One opinionated comment, otherwise nitts or questions.

policyengine/constants.py
anth-volk: I've refactored this to make more sense

PR #131, opened 2025-05-15
Enable cliff impacts

### LINKED ISSUE #130: Enable cliff impacts

nikhilwoodruff:
Fixes #130

### DIFF
--- a/policyengine/outputs/macro/comparison/calculate_economy_comparison.py
+++ b/policyengine/outputs/macro/comparison/calculate_economy_comparison.py
@@ -793,16 +793,21 @@ class EconomyComparison(BaseModel):
 
 def calculate_economy_comparison(
     simulation: Simulation,
+    include_cliffs: bool = False,
 ) -> EconomyComparison:
     """Calculate comparison statistics between two economic scenarios."""
     if not simulation.is_comparison:
         raise ValueError("Simulation must be a comparison simulation.")
 
-    baseline: SingleEconomy = simulation.calculate_single_economy(reform=False)
-    reform: SingleEconomy = simulation.calculate_single_economy(reform=True)
+    baseline: SingleEconomy = simulation.calculate_single_economy(
+        reform=False, include_cliffs=include_cliffs
+    )
+    reform: SingleEconomy = simulation.calculate_single_economy(
+        reform=True, include_cliffs=include_cliffs
+    )
     options = simulation.options
     country_id = options.country
-    if baseline.type == "general":
+    if not include_cliffs:
         budgetary_impact_data = budgetary_impact(baseline, reform)
         detailed_budgetary_impact_data = detailed_budgetary_impact(
             baseline, reform, country_id
@@ -839,7 +844,7 @@ def calculate_economy_comparison(
             labor_supply_response=labor_supply_response_data,
             constituency_impact=constituency_impact_data,
         )
-    elif baseline.type == "cliff":
+    elif include_cliffs:
         return dict(
             baseline=dict(
                 cliff_gap=baseline.cliff_gap,
--- a/policyengine/outputs/macro/comparison/calculate_economy_comparison.py
+++ b/policyengine/outputs/macro/comparison/calculate_economy_comparison.py
@@ -793,21 +793,16 @@ class EconomyComparison(BaseModel):
 
 def calculate_economy_comparison(
     simulation: Simulation,
-    include_cliffs: bool = False,
 ) -> EconomyComparison:
     """Calculate comparison statistics between two economic scenarios."""
     if not simulation.is_comparison:
         raise ValueError("Simulation must be a comparison simulation.")
 
-    baseline: SingleEconomy = simulation.calculate_single_economy(
-        reform=False, include_cliffs=include_cliffs
-    )
-    reform: SingleEconomy = simulation.calculate_single_economy(
-        reform=True, include_cliffs=include_cliffs
-    )
+    baseline: SingleEconomy = simulation.calculate_single_economy(reform=False)
+    reform: SingleEconomy = simulation.calculate_single_economy(reform=True)
     options = simulation.options
     country_id = options.country
-    if not include_cliffs:
+    if not simulation.options.include_cliffs:
         budgetary_impact_data = budgetary_impact(baseline, reform)
         detailed_budgetary_impact_data = detailed_budgetary_impact(
             baseline, reform, country_id
@@ -844,7 +839,7 @@ def calculate_economy_comparison(
             labor_supply_response=labor_supply_response_data,
             constituency_impact=constituency_impact_data,
         )
-    elif include_cliffs:
+    else:
         return dict(
             baseline=dict(
                 cliff_gap=baseline.cliff_gap,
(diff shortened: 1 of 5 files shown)


policyengine/outputs/macro/single/calculate_single_economy.py
```diff
@@ -414,7 +424,9 @@ def calculate_single_economy(
             "age": age,
             **labor_supply_responses,
             **lsr_working_hours,
-            "type": "general",
+            "type": "general" if include_cliffs else "cliff",
```
anth-volk, line 427: __minor, blocking:__ Is this not `"cliff" if include_cliffs else "general"`?

policyengine/outputs/macro/single/calculate_single_economy.py
```diff
@@ -331,6 +333,7 @@ def calculate_uk_programs(self) -> Dict[str, float]:
 def calculate_single_economy(
     simulation: Simulation, reform: bool = False
 ) -> Dict:
+    include_cliffs = simulation.options.include_cliffs
```
anth-volk, line 336: __question/suggestion:__ Is the structure here too muddled? And do we need all other single-economy calculations when running cliff impacts? I'd suggest breaking cliff impacts out into another calculation, as I believe occurs with API v1, unless all of the single-economy outputs are required in addition to the cliff impact outputs.

If you had a `calculate_cliff_impacts` method here that just handled that, then had a different return setup that more closely mirrored API v1, I think you'd find this code easier to parse and maintain, and I wonder if there'd be a slight bit of runtime savings?

policyengine/outputs/macro/single/calculate_single_economy.py
```diff
@@ -382,6 +385,13 @@ def calculate_single_economy(
         except:
             total_state_tax = 0
 
+    if include_cliffs:
```
anth-volk, line 388: __comment:__ If you were breaking these up into two separate calculation methods, this could basically just be your entire method body

policyengine/outputs/macro/comparison/calculate_economy_comparison.py
```diff
@@ -802,7 +802,7 @@ def calculate_economy_comparison(
     reform: SingleEconomy = simulation.calculate_single_economy(reform=True)
     options = simulation.options
     country_id = options.country
-    if baseline.type == "general":
+    if not simulation.options.include_cliffs:
```
anth-volk, line 805: __suggestion:__ I'd avoid this nested if treatment

Two reasons:
1. Clean code patterns recommend just returning from the if loop for readability
2. You could some day down the road have unintended consequences for an options setup that has `include_cliffs`, but also something else that you want to have branched logic for

policyengine/outputs/macro/comparison/calculate_economy_comparison.py
```diff
@@ -839,7 +839,7 @@ def calculate_economy_comparison(
             labor_supply_response=labor_supply_response_data,
             constituency_impact=constituency_impact_data,
         )
-    elif baseline.type == "cliff":
+    else:
```
anth-volk, line 842: __question, blocking:__ Are cliff impacts meant to be treated as an additional option within an economy, or a standalone calculation?

From what I'm reading here and in `calculate_single_economy.py`, when running a single-economy calculation, it's additive, while here, it's either/or

policyengine/outputs/macro/single/calculate_single_economy.py
```diff
@@ -49,6 +49,8 @@ class SingleEconomy(BaseModel):
     weekly_hours_substitution_effect: float | None
     type: str
```
anth-volk, line 50: __issue/suggestion:__ I've amended the type hint, but wonder if cliffs should just be completely removed here and you have two single-economy calculation types: `SingleEconomy` and something that only does cliffs. Curious to hear your thoughts.
```suggestion
    type: Literal["general", "cliff"]
```

policyengine/outputs/macro/single/calculate_single_economy.py
```diff
@@ -382,6 +385,13 @@ def calculate_single_economy(
         except:
             total_state_tax = 0
 
+    if include_cliffs:
+        cliff_gap = task_manager.simulation.calculate("cliff_gap")
+        is_on_cliff = task_manager.simulation.calculate("is_on_cliff")
+        total_cliff_gap = cliff_gap.sum()
+        total_adults = task_manager.simulation.calculate("is_adult").sum()
+        cliff_share = is_on_cliff.sum() / total_adults
+
```
anth-volk, line 394: __issue:__ Please type hint these

policyengine/outputs/macro/single/calculate_single_economy.py
nikhilwoodruff: Yes, thanks- fixed.

policyengine/outputs/macro/single/calculate_single_economy.py
nikhilwoodruff: I'd like to propose for this fix that we return cliff impacts along with everything else in the normal impacts, rather than as a separate endpoint. I think there are several justifications:

* There is less confusion over having separate endpoints for separate economy comparison simulations (cliff impacts was dropped in the first place because of how inconsistently it is handled from everything else)
* There is more consistency with other types of simulation options: everything is handled from the same function
* It's less code to include in the main function, so less code to maintain

We added cliff impacts a few years ago, and since then we've consolidated the 'simulation options' into a distinct set of options users select rather than an extra output type that, unlike anything else, re-triggers a simulation. I think in app-v2 we should definitely make 'include cliff impacts' a simulation option users select or don't select, but this is a fix that works in the meantime.

policyengine/outputs/macro/comparison/calculate_economy_comparison.py
nikhilwoodruff: Yes- made these routes additive.

policyengine/outputs/macro/single/calculate_single_economy.py
nikhilwoodruff: More detail above, but I think there is a case that adding it in here is cleaner.

policyengine/outputs/macro/single/calculate_single_economy.py
nikhilwoodruff: I should also add that there is no real performance gain to separating it out: cliff impacts require by definition household net income to be calculated for the microdata and is pretty much a superset of the requirements of the main economy comparison.

anth-volk (commented):
It was a real pleasure going through this code @nikhilwoodruff. Left some minor change recommendations, but think we're close to the finish line!

policyengine/outputs/macro/single/calculate_single_economy.py
```diff
@@ -414,7 +430,9 @@ def calculate_single_economy(
             "age": age,
             **labor_supply_responses,
             **lsr_working_hours,
-            "type": "general",
+            "type": "general" if not include_cliffs else "cliff",
             "programs": uk_programs,
+            "cliff_gap": float(total_cliff_gap) if include_cliffs else None,
```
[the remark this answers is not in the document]
nikhilwoodruff: It's unnecessary yes

anth-volk (approved):
LGTM! I'll open a separate issue in API to add a cliff impact test for integration testing purposes.

PR #132, opened 2025-05-15
Add CachingGoogleStorageClient

mikesmit:

This commit adds, but does not yet use, CachingGoogleStorageClient which is a class used to monitor a remote file in google storage for changes caching the result locally to disk.

### DIFF
--- a/tests/utils/data/test_caching_google_storage_client.py
+++ b/tests/utils/data/test_caching_google_storage_client.py
@@ -1,0 +1,107 @@
+from pathlib import Path
+import pytest
+import tempfile
+from unittest.mock import MagicMock, create_autospec, patch
+from policyengine.utils.data import CachingGoogleStorageClient
+from tests.utils.data.conftest import MockedStorageSupport
+
+
+class TestCachingGoogleStorageClient:
+    @pytest.mark.asyncio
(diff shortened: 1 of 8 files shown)


tests/utils/data/test_caching_google_storage_client.py
```diff
@@ -0,0 +1,137 @@
+from pathlib import Path
+import pytest
+import tempfile
+from unittest.mock import MagicMock, create_autospec, patch
+from policyengine.utils.data import CachingGoogleStorageClient
+
+
+class MockedStorageSupport:
```
anth-volk, line 8: __suggestion, non-blocking:__ Separate fixtures out into `test/fixtures/data/filepath`

anth-volk (approved):
Just had one minor suggestion, but otherwise this would be great to have, thanks @mikesmit

tests/utils/data/test_caching_google_storage_client.py
mikesmit: Chatted on slack and decided to put it in a conftest.py file. which I'm doing in a new revision now.

I'm kind of on the fence here. If it's small and only used by one test I tend to want to have it next to the thing being tested, but I also buy it's better to remove clutter and just have the test code in the test file.

PR #133, opened 2025-05-15
download_file_from_gcs now uses cache.

mikesmit:

this change re-implements the download_file_from_gcs function to use CachingGoogleStorageClient.

Files should now be cached locally on a per-process basis and only updated on disk when the remote version crc changes.

### DIFF
--- a/policyengine/utils/google_cloud_bucket.py
+++ b/policyengine/utils/google_cloud_bucket.py
@@ -1,3 +1,23 @@
+from .data.caching_google_storage_client import CachingGoogleStorageClient
+import asyncio
+from pathlib import Path
+
+_caching_client: CachingGoogleStorageClient | None = None
+
+
+def _clear_client():
+    global _caching_client
+    _caching_client = None
+
+
+def _get_client():
+    global _caching_client
+    if _caching_client is not None:
+        return _caching_client
+    _caching_client = CachingGoogleStorageClient()
+    return _caching_client
+
+
 def download_file_from_gcs(
     bucket_name: str, file_name: str, destination_path: str
 ) -> None:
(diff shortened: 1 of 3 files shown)


anth-volk (approved):
Per discussion with @mikesmit, due to lack of integration testing capabilities and the fact that the v2 service is not used by customers, we will merge and roll back on off chance that file downloads are misconfigured.

policyengine/utils/google_cloud_bucket.py
```diff
@@ -1,3 +1,26 @@
+from .data.caching_google_storage_client import CachingGoogleStorageClient
```
anth-volk, line 1: __issue, blocking:__ Add a changelog entry

Without it, this package won't launch to PyPI, and there's no check for `changelog_entry.yaml` in the GH Actions

policyengine/utils/google_cloud_bucket.py
```diff
@@ -1,3 +1,26 @@
...
+def _clear_client():
+    global _caching_client
+    print("Clearing caching client")
+    _caching_client = None
+
+
+def _get_client():
+    global _caching_client
+    if _caching_client is not None:
+        print("Caching client already created")
```
anth-volk, line 17: __question:__ Is there an advantage to printing over using the logging package?

policyengine/utils/google_cloud_bucket.py
```diff
@@ -1,3 +1,26 @@
...
+        return _caching_client
+    print("Creating caching client")
```
anth-volk, line 19: __question:__ Here, too, would it be better to log?

anth-volk (changes requested):
A couple minor questions and one blocking issue: a `changelog_entry.yaml` file is needed to trigger PyPI deployment

policyengine/utils/google_cloud_bucket.py
mikesmit: oh that was me debugging. Will remove this and the other one.

PR #136, opened 2025-05-16
Handle bad `enhanced_cps` specification in Simulation initialisation

### LINKED ISSUE #134: Handle bad `enhanced_cps` specification in Simulation initialisation
Some of our regression tests are failing because the request passes in a bad schema. While we may be somewhat confident that the current version of the APIv1 passes in a valid schema, we could add an extra check on the policyengine.py side to handle this case, since it's obvious what to do:

1. If `enhanced_us` is in the `region` field, the user wants to use that dataset.
2. Set the dataset to the enhanced CPS dataset and set the region to None.

### LINKED ISSUE #135: Allow 2023 as a simulation year
We currently disallow this.

nikhilwoodruff:
Fixes #134
Fixes #135

### DIFF
--- a/policyengine/simulation.py
+++ b/policyengine/simulation.py
@@ -176,7 +176,7 @@ class Simulation:
         self,
     ):
         options = self.options
-        if "enhanced_us" in options.region:
+        if options.region is not None and "enhanced_us" in options.region:
             options.region = None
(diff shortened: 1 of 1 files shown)


anth-volk:
@nikhilwoodruff sounds like we're going to explicitly forbid the use of this parameter in v2. Holding off on this PR in favor of changes to the API and/or app to just prevent usage of this.

PR #143, opened 2025-05-19
Reenable download from cache but not async

mikesmit:

this change re-implements the download_file_from_gcs function to use CachingGoogleStorageClient and removes the async stuff from the client to avoid issues with the existing event loop.

Files should now be cached locally on a per-process basis and only updated on disk when the remote version crc changes.

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    fixed:
+      - google storage caching is now fully sync, not async and reenabled.
--- a/policyengine/utils/google_cloud_bucket.py
+++ b/policyengine/utils/google_cloud_bucket.py
@@ -1,3 +1,23 @@
+from .data.caching_google_storage_client import CachingGoogleStorageClient
+import asyncio
+from pathlib import Path
+
+_caching_client: CachingGoogleStorageClient | None = None
+
+
+def _get_client():
+    global _caching_client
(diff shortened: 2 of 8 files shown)


changelog_entry.yaml
```diff
@@ -0,0 +1,5 @@
+=======
```
anth-volk, line 1: __issue, minor, blocking:__ Please remove this line

mikesmit: oops. Will do.

policyengine/utils/google_cloud_bucket.py
```diff
@@ -1,3 +1,23 @@
...
+
+def _get_client():
+    global _caching_client
+    if _caching_client is not None:
+        return _caching_client
+    _caching_client = CachingGoogleStorageClient()
+    return _caching_client
+
+
+def _clear_client():
```
anth-volk, line 16: __question:__ Is this function used anywhere? And is it meant to be?

mikesmit: Yes. In the unit test. Otherwise no.

anth-volk: Thanks, should've read through all changes before commenting

anth-volk (approved):
@mikesmit Left one minor blocking comment that I believe you already resolved. Otherwise, looks good to me!

PR #145, opened 2025-05-21
Use Google Cloud Storage by default

nikhilwoodruff:


policyengine/utils/data_download.py
```diff
@@ -27,18 +27,6 @@ def download(
     )
```
anth-volk, line 27: (This comment should be attached to the `DataFile` initialization above, GH review limitations disallowed)
__question:__ Is this method structure too limiting and envision ourselves as the only customer?

We've had quite a few issues around prioritization HF/Google Cloud data storage. In order to make it easier for ourselves and to make `.py` more open to any user, I'd suggest implementing `download()` with an OOP strategy pattern, passing in two args: a string literal corresponding to one permitted strategy, perhaps providing a default (e.g., "google_cloud_bucket"), then another arg called "options" that was one of a variety of options schemas, e.g., `gcs_bucket` for Google downloads and the two HF options for Hugging Face.

This would do the following:
* Enable us to easily override the default behavior wherever we want without being tied to a particular prioritization
* Enable us to build new download strategies as needed, e.g., if we ever hosted something on GitHub again or on BitBucket
* Enable external users to build their own download strategies and plug them in as needed

policyengine/utils/data_download.py
```diff
@@ -27,18 +27,6 @@ def download(
...
 
     logging.info = print
-    # NOTE: tests will break on build if you don't default to huggingface.
```
anth-volk, line 30: __question, blocking:__ Is the UK non-public data hosted in a bucket? I thought I remembered it being GCP-only

policyengine/utils/data_download.py
```diff
@@ -53,6 +41,19 @@ def download(
         )
         return filepath
 
+    # NOTE: tests will break on build if you don't default to huggingface.
```
anth-volk, line 44: __nit:__ This comment would no longer be true

anth-volk (commented):
@nikhilwoodruff had some thoughts for you around shifting how we do downloads to prevent future bugginess. Curious to hear your thoughts.

policyengine/utils/data_download.py
mikesmit: **suggestion, non-blocking** My 2 cents: I agree directionally. But right now I'd simplify. Remove the hugging face entirely. The first customer is simulation service and that's all we need.

Get storage & caching working rock solid and usable from dev desktop, pipeline, and service. Then expand out.

policyengine/utils/data_download.py
mikesmit: Agree.

policyengine/utils/data_download.py
```diff
@@ -27,18 +27,6 @@ def download(
     )
 
     logging.info = print
```
mikesmit, line 29: **Suggestion** While we're in here, could this use the Python logger please? I'd make this blocking, but the switch to gcp is probably more important.

mikesmit (commented):
Let's discuss why hugginface at all. otherwise let's go.

nikhilwoodruff:
To do: just don't do huggingface in policyengine.py

anth-volk:
Closing, as my understanding is that this is superseded by #148. Feel free to reopen if I'm wrong @nikhilwoodruff.

PR #146, opened 2025-05-22
Always check for a new data file version even if one has been downloa…

mikesmit:
…ded.

The existing code is pretty inconsistent in terms of how/when it decides to try to download a data file and we haven't clearly defined the intended behavior.

We are prioritizing the simulation API use case in which case we always want to use the most recent version of the data file for a simulation.

This change means that if the code specifies a remote data file (either by explicitly giving a url or by defaulting to a country dataset) we will always check for a new version when creating a Simulation object even if we have a local copy.

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    fixed:
+    - Always look for new data file versions even if we have a local copy of one.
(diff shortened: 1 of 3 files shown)


anth-volk (approved):
LGTM! Would appreciate holding off on merging until May 23 while I run some regression tests locally.

PR #152, opened 2025-05-27
Use a persistent location for caching data files

### LINKED ISSUE #151: Use a persistent location for caching data files
Right now the cache we use for data is reset on termination of the python process. This means data downloads are repeated every time a process is restarted. We could instead set a persistent location for the cache (since we're making sure we don't reuse outdated data by checking CRCs) so that we can avoid repeated downloads.

nikhilwoodruff:
Fixes #151

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    fixed:
+    - Persistent cache location for data files.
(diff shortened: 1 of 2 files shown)


mikesmit (commented):
Unless there is a really strong benefit I would not bother with this right now. 

Actually, side note, if we had integrated metrics in policyengine.py we could easily see how frequently we're downloading and how long it takes.

I had mentioned "instrumentation" as a key requirement for policyengine.py a while back this is a great example where I wanted to add it but didn't have a mechanism.

PR #165, opened 2025-06-16
Add option to reach data download token from b64 SA keys

### LINKED ISSUE #164: Add option to reach data download token from b64 SA keys.
We should be able to give users access tokens for the microdata.

nikhilwoodruff:
Fixes #164

After this PR, we could give users access tokens for private microdata. The process would be:

* Create a service account and give it the permissions `Storage legacy bucket reader` and `Storage legacy object reader` in the relevant Google Cloud bucket in the project `PolicyEngine research`
* Generate a JSON private key and base64-encode the contents
* Pass this key to a client

The argument for b64 encoding is that it reduces complexity and bug-prone behaviour around string formatting etc. when pasting the contents.

### DIFF
--- a/policyengine/utils/data/simplified_google_storage_client.py
+++ b/policyengine/utils/data/simplified_google_storage_client.py
@@ -2,7 +2,11 @@ import asyncio
 from policyengine_core.data.dataset import atomic_write
 import logging
 from google.cloud.storage import Client, Blob
+from google.oauth2 import service_account
 from typing import Iterable, Optional
+import os
+import json
+import base64
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


policyengine/utils/data/simplified_google_storage_client.py
```diff
@@ -16,7 +20,18 @@ class SimplifiedGoogleStorageClient:
     """
 
     def __init__(self):
```
anth-volk, line 22: __question, blocking:__ Have you looked at all into Google Secrets Manager?

An option I've found that we could use would be, instead of having us create individual service accounts for customers, we could generate our own API keys and then use Google Secrets Manager to store these keys and pass & validate them via SHA256 encryption. This seems a bit more akin to what you're seeking to do. It also gives us more granular control over which datasets this enables, as we can store richer metadata with the key (imagine one day we create another type of limited-access dataset that we don't want these customers to access).

I have a Claude chat [here](https://claude.ai/artifacts/7ebbf51a-b952-4320-b256-625c3b80233b) describing Google Secrets Manager, as well as other options that may be less desirable (e.g., JWT tokens). The most relevant portions are toward the end. Curious to hear your thoughts.

anth-volk (commented):
@nikhilwoodruff Question for you around desired implementation.

policyengine/utils/data/simplified_google_storage_client.py
mikesmit: I think I need to understand our use case better here. I don't really want to manage service accounts for external users or force the user to authenticate using a specific mechanism.

Is there a reason we can't have them provide a gmail account or service account that they own/manage and then grant it permission to access the bucket?

nikhilwoodruff: The reason is that we want to keep the process of running code that downloads the microdata simple. A single microdata access token as an env var is as simple as it gets

PR #175, opened 2025-08-13
Fix UK dataset issue

### LINKED ISSUE #176: UK dataset issue with FRS 2023-24
The UK dataset had an issue that needed fixing when moving to the FRS 2023-24 dataset. This involved updating the dataset references and ensuring compatibility with the new fiscal year data.

nikhilwoodruff:
Fixes #176 

Going to merge on tests pass but flagging for awareness @anth-volk 

I had to adjust the tests:

* Use the FRS rather than a sample dataset (this actually feels safer to me)
* The new UK dataset format doesn't have the file path saved so we can't test it (but this is fine)
* The `called_with` functionality seems to be dropped in 3.13

Web app currently broken and these tests didn't catch a far more important failure so not going to block on improving them here.


nikhilwoodruff:
Had to remove the single household calculation page because of [this issue](https://github.com/PolicyEngine/policyengine-us/issues/6400). policyengine.py's household features are not used in production so not a blocker.

PR #177, opened 2025-08-22
Add database models

nikhilwoodruff:
cc @MaxGhenis 

Removed the actual calculation logic only temporarily- will add back in but as a function of the db rather than simulations.


nikhilwoodruff:
@anth-volk now ready for review. This has the UK basics but I'll add the rest shortly. We use the Pydantic BaseModels in general, but I've added SQLmodel equivalents and a Database class that users can then use to upload/get from the database.

src/policyengine/models/dynamics.py
```diff
@@ -0,0 +1,33 @@
...
+    parent_dynamics: Dynamics | None = None
+
+    # Metadata
+    description: str | None = None
+    created_at: datetime | None = None
+    # Should automatically be current datetime
+    updated_at: datetime | None = None
+
+    parameter_values: list[ParameterValue] | None = None
+    simulation_modifier: Any | None = None
```
anth-volk, line 32: __question:__ My understanding is that a `simulation_modifier` is a func. How does that work here?

src/policyengine/models/enums.py
```diff
@@ -0,0 +1,19 @@
...
+class OperationStatus(str, Enum):
+    """Status of simulation processing."""
+
+    PENDING = "pending"
+    RUNNING = "running"
+    COMPLETED = "completed"
+    FAILED = "failed"
+
+
+class DatasetType(str, Enum):
```
anth-volk, line 15: __suggestion, minor:__ Move this to the `Dataset` type file

src/policyengine/models/parameter.py
```diff
@@ -0,0 +1,47 @@
...
+"""
+
+from __future__ import annotations
+
+from datetime import datetime
+from typing import TYPE_CHECKING, Any
+
+from pydantic import BaseModel
+
+if TYPE_CHECKING:  # Import to satisfy type checkers without runtime cycles
```
anth-volk, line 13: __question:__ What does the point about runtime cycles mean?

src/policyengine/models/parameter.py
```diff
@@ -0,0 +1,47 @@
...
+    """Policy parameter of the country package model."""
+
+    name: str
+    parent: Parameter | None = None  # For hierarchical parameters
+
+    # Parameter metadata
+    label: str | None = None
+    description: str | None = None
+    unit: str | None = None
+    data_type: type | None  # "float", "int", "bool", "string"
```
anth-volk, line 28: __question:__ What about enums? I think this'd need something to hold the enum options?

src/policyengine/models/parameter.py
```diff
@@ -0,0 +1,47 @@
...
+    """Individual parameter value for some point in time."""
+
+    # Foreign keys
+    policy: Policy | None = None
+    dynamics: Dynamics | None = None
+    parameter: Parameter
+    model_version: str
+
+    # Time period for this change
+    start_date: datetime
```
anth-volk, line 42: __question:__

In JS, all dates and times are written to a particular time zone, which causes annoying bugs if you write something that isn't fixed to one zone (e.g., UTC). Are these `datetime` instances zoned? If not, can they be? Alternatively, is it possible to write ISO date strings (YYYY-MM-DD) and them still be stored as dates?

src/policyengine/models/parameter.py
```diff
@@ -0,0 +1,47 @@
...
+    end_date: datetime | None = None
+
+    # The actual change
```
anth-volk, line 45: __nit:__ What does this comment mean?

src/policyengine/models/parameter.py
```diff
@@ -0,0 +1,47 @@
...
+    value: Any  # JSON-serializable value
+    country: str | None = None
```
anth-volk, line 47: __suggestion:__ Make this `country_id` and fix it to a set options list

src/policyengine/models/policy.py
```diff
@@ -0,0 +1,26 @@
...
+    """Modifications made to baseline tax-benefit rules."""
+
+    name: str | None = None
+
+    # Metadata
+    description: str | None = None
+
+    parameter_values: list[ParameterValue] | None = None
+    simulation_modifier: Any | None = None
+    country: str | None = None
```
anth-volk, line 26: Same questions as in `models/parameter`

src/policyengine/models/reports.py
```diff
@@ -0,0 +1,98 @@
...
+    """An element of a report composed in the web app.
+
+    Subclasses represent concrete charts/tables. They implement `run()` and
+    return a list of data records (BaseModels). Callers can convert that list
+    to a DataFrame using the record class' helper.
+    """
+
+    name: str | None = None
+    description: str | None = None
+    report: Optional["Report"] = None
```
anth-volk, line 29: __question:__ Is this written this way to allow cleaner initialization/pre-creation of empty objects?

src/policyengine/models/reports.py
```diff
@@ -0,0 +1,98 @@
...
+    group_value: int | str
+    entity_level: str = "household"
+    time_period: int | str | None = None
+
+    total_change: float
+    relative_change: float
+    average_change_per_entity: float
+
+class VariableChangeGroupByQuantileGroup(ReportElementDataItem):
+    """Winners/losers share by quantile grouping based on baseline.
```
anth-volk, line 68: __suggestion:__ Somewhere (maybe the docs, not here), break down what each of these report output types are meant to be used for and how some of the keys are meant to function.

src/policyengine/models/user.py
```diff
@@ -0,0 +1,37 @@
+from pydantic import BaseModel
+from typing import Optional
+from uuid import UUID
+
+
+class User(BaseModel):
+    id: UUID
+    name: str
+    email: str
```
anth-volk, line 9: __question:__ Do we want this as required? Our users don't currently have to have an email

src/policyengine/models/user.py
```diff
@@ -0,0 +1,37 @@
...
+class UserReport(BaseModel):
+    user: User
+    report: "Report"
+    label: Optional[str] = None
+    description: Optional[str] = None
+
+
+class UserReportElement(BaseModel):
+    user: User
+    report_element: "ReportElement"
```
anth-volk, line 35: __question:__ Is this record to enable many-to-manys on reports and report elements?

src/policyengine/models/variable.py
```diff
@@ -0,0 +1,21 @@
...
+class Variable(BaseModel):
+    """PolicyEngine variable concept- an attribute of an entity."""
+
+    name: str
+
+    # Variable metadata
+    label: str | None = None
+    description: str | None = None
+    unit: str | None = None
+    data_type: type  # "float", "int", "bool", "string", "enum"
```
anth-volk, line 18: __question:__ Here again, would we need to store enum options?

src/policyengine/models/variable.py
```diff
@@ -0,0 +1,21 @@
...
+    entity: str  # "person", "household", "tax_unit", etc.
+    definition_period: str | None = None  # "year", "month", "eternity"
```
anth-volk, line 20: __suggestion:__ Constrain type to template literal

src/policyengine/report_elements/single_simulation.py
```diff
@@ -0,0 +1,106 @@
...
+            return []
+        else:
+            # Simulation.result is a Dataset with .data being SingleYearDataset
+            try:
+                data = sim.result.data  # type: ignore[attr-defined]
+            except Exception:
+                # If result is None or not shaped as expected
+                return []
+        if not isinstance(data, SingleYearDataset):
+            return []
```
anth-volk, line 47: __question:__ I'm surprised these fail silently

src/policyengine/report_elements/single_simulation.py
```diff
@@ -0,0 +1,106 @@
...
+
+        # Pick target table
+        table = data.tables.get(self.entity_level)
+        if table is None:
+            # fallback: find first table containing variable
+            for tbl in data.tables.values():
+                if self.variable in tbl.columns:
+                    table = tbl
+                    break
```
anth-volk, line 56: __suggestion:__ Feels like this should raise, as this fallback could create hard-to-find "bugs" that are just this fallback functioning correctly

src/policyengine/report_elements/single_simulation.py
```diff
@@ -0,0 +1,106 @@
...
+        if table is None or self.variable not in table.columns:
+            return []
+
+        # Use MicroDataFrame for weighted operations
+        use_weights = "weight_value" in table.columns
+
+        def compute_metric_from_df(df: pd.DataFrame) -> float:
+            mdf = MicroDataFrame(df, weights="weight_value") if use_weights else MicroDataFrame(df)
+            s = mdf[self.variable]
+            if self.metric == AggregateMetric.MEAN:
```
anth-volk, line 66: __accolade:__ Like the use of the enums here

src/policyengine/report_elements/two_simulation_comparison.py
```diff
@@ -0,0 +1,229 @@
...
+from policyengine.models.simulation import Simulation
+from policyengine.models.single_year_dataset import SingleYearDataset
+
+
+def _get_table(sim: Simulation, entity_level: str) -> pd.DataFrame | None:
+    try:
+        data = sim.result.data  # type: ignore[attr-defined]
+    except Exception:
+        return None
+    if not isinstance(data, SingleYearDataset):
```
anth-volk, line 24: __question:__ How will this work with US sims?

src/policyengine/report_elements/two_simulation_comparison.py
```diff
@@ -0,0 +1,229 @@
...
+        return None
+    return data.tables.get(entity_level)
+
+
+def _align_baseline_reform(
```
anth-volk, line 29: __suggestion:__ Doc why this is needed

src/policyengine/report_elements/two_simulation_comparison.py
```diff
@@ -0,0 +1,229 @@
...
+    variable: str
+    group_variable: str
+    group_value: Any
+    entity_level: str = "household"
+
+    def run(self) -> list[ChangeByBaselineGroup]:
+        base_tbl = _get_table(self.baseline, self.entity_level)
+        ref_tbl = _get_table(self.reform, self.entity_level)
+        if base_tbl is None or ref_tbl is None:
+            return []
```
anth-volk, line 66: __suggestion:__ Here, too (and I'm guessing in other places), I'm wondering if we should be throwing. This seems like it could be a pretty big issue.

src/policyengine/report_elements/two_simulation_comparison.py
```diff
@@ -0,0 +1,229 @@
...
+        mdf = MicroDataFrame(sub, weights=weights_col) if weights_col else MicroDataFrame(sub)
+        total_change = float(mdf["__delta__"].sum())
+        denom = float(mdf[f"{self.variable}_base"].sum())
+        relative_change = float(total_change / denom) if denom != 0 else float("nan")
+        # Sum of weights via summing a column of ones
+        sub = sub.copy()
+        sub["__ones__"] = 1.0
+        mdf2 = MicroDataFrame(sub, weights=weights_col) if weights_col else MicroDataFrame(sub)
+        weights_sum = float(mdf2["__ones__"].sum())
+        avg_change = float(total_change / weights_sum) if weights_sum != 0 else float("nan")
```
anth-volk, line 92: __suggestion:__ Refactor into one or more methods

src/policyengine/report_elements/two_simulation_comparison.py
```diff
@@ -0,0 +1,229 @@
...
+        ):
+            return []
+
+        merged = _align_baseline_reform(base_tbl, ref_tbl, self.entity_level)
+        merged["__delta__"] = merged[f"{self.variable}_ref"] - merged[f"{self.variable}_base"]
+        if self.change_bound_is_relative:
+            base = merged[f"{self.variable}_base"].replace({0.0: np.nan}).abs()
+            merged["__delta__"] = merged["__delta__"] / base
+
+        results: list[VariableChangeGroupByQuantileGroup] = []
```
anth-volk, line 141: I'm guessing there's some overlap with the above record, so if you had some util functions, the length of this file would be quite a bit shorter. Also, if we ever modify something around weight calculations, maintenance will be less annoying.

src/policyengine/tables/dataset.py
```diff
@@ -0,0 +1,21 @@
...
+from sqlalchemy import LargeBinary
+from sqlmodel import SQLModel, Field
+
+from policyengine.models.enums import DatasetType
+
+
+class DatasetTable(SQLModel, table=True):
+    __tablename__ = "datasets"
+
+    id: UUID = Field(default_factory=uuid4, primary_key=True)
```
anth-volk, line 15: __question:__ Just so I know for when we use it elsewhere, do we intend to migrate to UUIDs for all ID types?

src/policyengine/tables/dynamics.py
```diff
@@ -0,0 +1,21 @@
...
+class DynamicsTable(SQLModel, table=True):
+    __tablename__ = "dynamics"
+
+    id: UUID = Field(default_factory=uuid4, primary_key=True)
+    name: str | None = None
+    parent_id: UUID | None = Field(default=None, foreign_key="dynamics.id")
+    description: str | None = None
+    created_at: datetime | None = None
+    updated_at: datetime | None = None
+    country: str | None = None
```
anth-volk, line 19: __question:__ Can this be an enum with set options? Would be neat for the tables to constrain in that way.

src/policyengine/tables/reports.py
```diff
@@ -0,0 +1,32 @@
...
+
+class ReportTable(SQLModel, table=True):
+    __tablename__ = "reports"
+
+    id: UUID = Field(default_factory=uuid4, primary_key=True)
+    name: str
+    description: str | None = None
+    country: str | None = None
+
+    # No ORM relationship wiring
```
anth-volk, line 19: __question:__ Why?

src/policyengine/utils/metadata.py
```diff
@@ -0,0 +1,106 @@
...
+        "parameters": parameters,
+        "parameter_values": parameter_values,
+        "variables": variables,
+    }
+
+def _safe_date(date_str: str | None) -> datetime.datetime | None:
+    """Parse a date safely.
+
+    Accepts YYYY-MM-DD and ISO timestamps (YYYY-MM-DDTHH:MM:SS), and treats
+    "infinity" / "+infinity" or None as open-ended (returns None).
```
anth-volk, line 81: __question/issue:__ What about `-Infinity`? Also, this code doesn't handle the JS standard, which uses a capital "I", and which is what is written somewhere else in this PR.

anth-volk: __separate issue:__ I see now that `-infinity`/`-inf` are handled, but they're treated as None. Imagine you have a policy that's from `-inf` to tomorrow. This will cause a start date of `None`, which violates the `Parameter` model and table schemas.

src/policyengine/countries/uk/simulation.py
```diff
@@ -0,0 +1,72 @@
...
+        if simulation.dynamics.simulation_modifier:
+            simulation.dynamics.simulation_modifier(sim)
+
+    scenario = PolicyEngineUKScenario(simulation_modifier=modifier)
+    sim = PolicyEngineUKSimulation(
+        dataset=PolicyEngineUKSingleYearDataset(
+            person=simulation.dataset.data.tables["person"],
+            benunit=simulation.dataset.data.tables["benunit"],
+            household=simulation.dataset.data.tables["household"],
+            fiscal_year=simulation.dataset.data.year,
```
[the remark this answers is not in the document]
nikhilwoodruff: Dataset has 'dataset type' built for this purpose: simulation.run() should check for that and run whatever operation is needed.

src/policyengine/countries/uk/simulation.py
```diff
@@ -0,0 +1,72 @@
...
+    # Implement UK-specific simulation logic here
+    sim = _get_simulation(simulation)
+
+    output = simulation.dataset.data.copy()
+    output.tables["household"]["gov_tax"] = sim.calculate(
+        "gov_tax",
+    )
+    output.tables["household"]["gov_spending"] = sim.calculate("gov_spending")
+    output.tables["household"]["gov_balance"] = sim.calculate("gov_balance")
+
```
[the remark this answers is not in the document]
nikhilwoodruff: How do you mean handlers? If it's just to submodule when this gets bigger, yes

src/policyengine/countries/us/datasets.py
```diff
@@ -0,0 +1,212 @@
...
+from typing import Dict
+
+import pandas as pd
+
+from policyengine.models.dataset import Dataset
+from policyengine.models.single_year_dataset import SingleYearDataset
+from policyengine.models.enums import DatasetType
+
+
+def create_ecps(year: int = 2024) -> Dataset:
```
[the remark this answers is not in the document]
nikhilwoodruff: These are actually needed for right now so we have a way to put new datasets in the database. In future we'll automate this by having the data packages publish straight into the database.

src/policyengine/countries/us/simulation.py
```diff
@@ -0,0 +1,84 @@
...
+        simulation.dynamics.parameter_values or []
+    )
+
+    def modifier(sim):
+        parametric_policy.simulation_modifier(sim)
+        if simulation.policy.simulation_modifier:
+            simulation.policy.simulation_modifier(sim)
+
+        parametric_dynamics.simulation_modifier(sim)
+        if simulation.dynamics.simulation_modifier:
```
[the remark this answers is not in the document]
nikhilwoodruff: Nope, each Policy object has a structural reform option: just set the simulation_modifier attribute. It's lost when put into the database though since afaict there is no good way to store the python function in the database and retrieve it

pyproject.toml
```diff
@@ -66,3 +75,23 @@ extend-exclude = '''
   | dist
 )/
 '''
+
+[tool.ruff]
+line-length = 79
+target-version = "py311"
```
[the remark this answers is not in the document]
nikhilwoodruff: Good catch

src/policyengine/countries/uk/datasets.py
```diff
@@ -0,0 +1,179 @@
...
+    # Drop any missing tables to avoid serialising Nones
+    tables = {k: v for k, v in tables.items() if v is not None}
+    data = SingleYearDataset(
+        tables=tables,
+        year=year,
+    )
+    return Dataset(name="efrs", data=data, dataset_type=DatasetType.UK)
+
+
+def create_uk_synthetic(year: int = 2029) -> Dataset:
```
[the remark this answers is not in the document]
nikhilwoodruff: Yes, removed

src/policyengine/countries/uk/simulation.py
```diff
@@ -0,0 +1,72 @@
...
+        parametric_dynamics.simulation_modifier(sim)
+        if simulation.dynamics.simulation_modifier:
+            simulation.dynamics.simulation_modifier(sim)
+
+    scenario = PolicyEngineUKScenario(simulation_modifier=modifier)
+    sim = PolicyEngineUKSimulation(
+        dataset=PolicyEngineUKSingleYearDataset(
+            person=simulation.dataset.data.tables["person"],
+            benunit=simulation.dataset.data.tables["benunit"],
+            household=simulation.dataset.data.tables["household"],
```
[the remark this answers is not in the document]
nikhilwoodruff: Yes- we can add shortcuts but better to preserve this structure for consistency

docs/models.md
```diff
@@ -6,9 +6,9 @@ These are:
 * `Parameter`: A parameter represents some global variable in the tax-benefit system. For example, the personal allowance in the UK income tax system is a parameter.
 * `ParameterValue`: A parameter value represents a specific value for a parameter at a given range in time (e.g. the personal allowance between April 2023 and April 2024 is £12,570).
 * `Policy`: A policy represents a set of `ParameterValue`s that together define a specific change to the tax-benefit system. For example, a policy might represent a change to the personal allowance, or a change to the rate of VAT.
-* `Dynamics`: A dynamics represents a set of rules that define how the tax-benefit system changes over time. For example, a dynamics might represent the annual uprating of parameters in the tax-benefit system.
+* `Dynamic`: A dynamic represents a set of rules that define how the tax-benefit system changes over time. For example, a dynamic might represent the annual uprating of parameters in the tax-benefit system.
```
MaxGhenis, line 9: Uprating as a dynamic? This is policy at least in the US. I think of dynamics as the feedbacks in the economy. We might want a separate term/concept for policy uncertainty.

src/policyengine/tables/dataset.py
nikhilwoodruff: Yes I think that would be good

src/policyengine/models/report_items/two_sim_change.py
```diff
@@ -0,0 +1,351 @@
...
+                # Resolve group column name and subset
+                if grp_var in merged.columns:
+                    grp_col = grp_var
+                elif f"{grp_var}_base" in merged.columns:
+                    grp_col = f"{grp_var}_base"
+                elif f"{grp_var}_ref" in merged.columns:
+                    grp_col = f"{grp_var}_ref"
+                else:
+                    grp_col = None
+
```
anth-volk, line 320: __question:__ Maybe this could also be something reusable?

tests/test_database_constant.py
```diff
@@ -0,0 +1,41 @@
...
+    # Make sure password is not set
+    old_password = os.environ.pop("POLICYENGINE_DB_PASSWORD", None)
+
+    try:
+        with pytest.raises(ValueError, match="POLICYENGINE_DB_PASSWORD"):
+            db = Database(POLICYENGINE_DB)
+    finally:
+        # Restore password if it was set
+        if old_password:
+            os.environ["POLICYENGINE_DB_PASSWORD"] = old_password
```
anth-volk, line 25: __issue, blocking:__ I'd strongly suggest against a test that pops environment variables out of a user's environment

anth-volk (commented):
Left a few extra comments, but all are targeted to specific things. Fully aligned on the broad direction.

src/policyengine/countries/us/metadata.py
```diff
@@ -0,0 +1,25 @@
...
+    US_AVAILABLE = False
+    system = None
+
+from policyengine.utils.metadata import get_metadata
+from policyengine.countries.us.datasets import create_ecps_years
+
+
+def get_us_metadata():
+    """Return US model metadata (variables, parameters, values) only."""
+    if not US_AVAILABLE:
```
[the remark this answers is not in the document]
nikhilwoodruff: Yes and also none- e.g. the main API shouldn't need to have the country packages installed.

PR #212, opened 2026-01-13
fix: clear caches after parameter updates in simulation modifier

nikhilwoodruff:
## Summary

Fixes an issue where policy reforms applied via `simulation_modifier_from_parameter_values` had no effect because the simulation was using cached values computed before the modifier ran.

The modifier now clears:
1. Parameter caches via `tax_benefit_system.reset_parameter_caches()`
2. Variable holder caches via `holder.delete_arrays()` for all populations

## Test plan

- [x] Added `tests/test_parametric_reforms.py` which verifies UC standard allowance changes take effect
- [x] Manually tested via policyengine-api-v2 `calculate_household_impact` function

### DIFF
--- a/src/policyengine/utils/parametric_reforms.py
+++ b/src/policyengine/utils/parametric_reforms.py
@@ -81,6 +81,18 @@ def simulation_modifier_from_parameter_values(
                 start=start_period,
                 stop=stop_period,
             )
+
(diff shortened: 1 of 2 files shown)


nwoodruff-co:
not sure about this- particularly the variable deletion. needs checking

PR #280, opened 2026-04-17
Bump pinned country-model versions to 3.9-compatible releases

MaxGhenis:
## Summary

Bump the stale pinned versions in the `[us]`, `[uk]`, and `[dev]` extras:

- `policyengine-us==1.602.0` → `==1.647.0` (first release with 3.9 support + breakdown fixes)
- `policyengine-uk==2.74.0` → `==2.88.0` (first release with 3.9 support)
- `policyengine_core>=3.23.6` → `>=3.24.1` (includes ParameterScale fix + 3.9 support)

## Motivation

The post-merge push workflow for #278 failed in the `Test (3.13)` and `Test (3.14)` jobs with:

```
ValueError: Parameter gov.contrib.additional_tax_bracket.bracket.rates
  has children ['8'] that are not in the possible values of the
  breakdown variable 'range(1, 8)'.
```

The stale `policyengine-us==1.602.0` and `policyengine-uk==2.74.0` pins no longer install cleanly under modern `policyengine-core` because the strict breakdown validator added in core 3.24.0 rejects older country-model parameter data. The fixes landed in PolicyEngine/policyengine-us#8045, #8049, #8051 and PolicyEngine/policyengine-uk#1625, so the minimum-version bump is enough to unblock the publish.

## Test plan

- [ ] CI `Test (3.13)` and `Test (3.14)` pass (they were the failing jobs post-merge)
- [ ] After merge, push workflow's Publish job runs → `policyengine 3.4.6` (or next) lands on PyPI with 3.9 support


MaxGhenis:
## Status update + scope reassessment

This PR was originally a surgical pin bump to unblock post-merge publish after PR #278 landed. The root failure was that `policyengine-us==1.602.0` / `policyengine-uk==2.74.0` were getting breakdown validation errors under the stricter core 3.24.x validator.

Once the pins are bumped to versions that include the breakdown fixes (us 1.647.0, uk 2.88.0) and core is pinned to 3.24.1+, the original blockers are resolved — `Install + smoke-import` now passes on all six Python versions (3.9–3.14), plus Lint, Mypy, changelog, and all docs/build jobs.

**But** pypkg's `Test (3.13)` / `Test (3.14)` integration tests surface two unrelated, pre-existing model-behavior issues that the new country-model versions expose:

### 1. US `RecursionError` on `refundable_ctc`

```
Exception: RecursionError while calculating refundable_ctc for period 2024. The full computation stack is:
  - income_tax 2024, itemizing
  - income_tax_before_refundable_credits 2024, itemizing
  - income_tax_capped_non_refundable_credits 2024, itemizing
  - income_tax_non_refundable_credits 2024, itemizing
  - non_refundable_ctc 2024, itemizing
  - refundable_ctc 2024, itemizing
```

Classic circular-formula loop: `refundable_ctc` is capped at `income_tax`, but `income_tax` needs `refundable_ctc` to compute net liability. Likely always present in the US model; just newly visible because pypkg's integration tests exercise the `itemizing` reform branch against us 1.647.0. Needs a us-side fix (either break the cycle or raise `max_spiral_loops` for this path).

### 2. UK `assert 0.0 > 0` in household-impact tests

`TestUKHouseholdImpact.test_single_adult_with_employment_income` and `.test_family_with_children` assert positive tax / child-benefit values. Under uk 2.88.0 they come back as `0.0`. Either the expected behaviour changed between uk 2.74 and 2.88, or there's a calculation regression. Needs a uk-side diagnosis.

## Recommendation

These are genuine upstream bugs in us / uk that happened to surface here, not issues with the pin bump itself. Rather than let the pin-bump PR grow into a debugging expedition:

- **Option A** — merge this PR with the two test files (`test_household_impact.py`, `test_us_reform_application.py`) marked as `pytest.mark.skip(reason="blocked on PolicyEngine/policyengine-us#NNNN")` (or equivalent) with explicit tracking issues. pypkg publishes again; the skipped tests serve as reminders.
- **Option B** — close this PR, keep pypkg pinned to the stale `1.602.0` / `2.74.0` versions, and live with a broken post-merge publish pipeline until someone fixes the us / uk issues and a coordinated bump happens.

Leaning toward (A). Happy to defer to maintainer judgement.

MaxGhenis:
Superseded by #282, which pairs the pin bumps here with the manifest-check relaxation needed to avoid the same blocker recurring.

Issue #258, opened 2026-03-17
Fix economic_impact_analysis() — correct programs, add budget summary, intra-decile, poverty demographics


anth-volk:
## Problem

`economic_impact_analysis()` in both `us/analysis.py` and `uk/analysis.py` is broken and untested. The API (`policyengine-api-v2-alpha`) bypasses it entirely, reimplementing all orchestration in `computation_modules.py` (818 lines) with workarounds for library bugs.

### What's broken today

| Issue | Detail |
|---|---|
| Wrong US program names | `payroll_tax` → `employee_payroll_tax`, `state_income_tax` not a program |
| Wrong US entity for `ssi` | Uses `person`, should be `spm_unit` |
| Too many US programs | 11 programs; API only uses 6 that work |
| Missing `model_rebuild()` | `ProgrammeStatistics` (UK) never rebuilt |
| `calculate_decile_impacts()` creates new sims | Wasteful when caller already has run simulations |
| No intra-decile computation | API computes this; library skips it |
| No budget summary | No class exists; API has workaround for double-weighting |
| No poverty by demographics | API computes age/gender/race breakdowns; library skips them |
| `StopIteration` on missing variables | Crashes instead of clear error |

## Solution

- Add `CountryConfig` strategy class with `US_CONFIG`/`UK_CONFIG` — eliminates country conditionals
- Add `compute_budget_summary()`, `compute_decile_impacts()`, `compute_program_statistics()` as individual functions
- Add `PolicyReformAnalysis` unified result container
- Rewrite `economic_impact_analysis()` for both countries to use individual compute functions
- Fix `model_rebuild()` in both `__init__.py` files
- Fix error handling in `Aggregate`/`ChangeAggregate` (`StopIteration` → `ValueError`)

This is Phase 1 of consolidating logic from `policyengine-api-v2-alpha` back into `policyengine.py`.

MaxGhenis:
Status check against main (2026-07-08 review): three of this issue's four asks have merged —

- correct US program mappings: #327
- intra-decile income changes: `outputs/intra_decile_impact.py`, wired into both countries' `economic_impact_analysis`
- poverty demographics: `outputs/poverty.py` (`calculate_us_poverty_by_age/gender/race`, UK age/gender)

[...]

PR #296, opened 2026-04-19
Add federal vs. state budgetary impact to economic_impact_analysis

### LINKED ISSUE #289: Add federal vs. state budgetary impact to economic_impact_analysis
## Context

PolicyEngine/policyengine-us#8076 added Person-level variables `federal_benefit_cost` / `state_benefit_cost` that split benefit expenditures between levels of government (currently covers Medicaid FMAP and CHIP eFMAP; will extend to SNAP OBBBA FY2028 state match, etc.).

Today `economic_impact_analysis()` in `src/policyengine/tax_benefit_models/us/analysis.py` returns per-program stats and aggregate impacts, but doesn't partition benefit spending by federal vs. state. Every consumer — policyengine-api's `compare.py::budgetary_impact` (PolicyEngine/policyengine-api#3481), analysis notebooks (PolicyEngine/analysis-notebooks#131), ad-hoc scripts — needs the same arithmetic, so the logic belongs here, not duplicated in each caller.

## Proposal

Extend `PolicyReformAnalysis` to expose:

- `federal_budgetary_impact`: federal tax revenue change (income_tax + payroll_tax) minus federal benefit spending change (`federal_benefit_cost`)
- `state_budgetary_impact`: state tax revenue change (state_income_tax) minus state benefit spending change (`state_benefit_cost`)
- `total_budgetary_impact`: existing sum, unchanged

Implementation: add two `ChangeAggregate(variable="federal_benefit_cost", type=ChangeAggregateType.SUM)` / `state_benefit_cost` calls alongside the existing per-program aggregates, and expose the federal/state splits as top-level fields on the analysis object.

Also tag `ProgramStatistics` entries with a `level_of_government` field (`"federal"`, `"state"`, or `"shared"`) so consumers can filter per-program rows. For shared programs like Medicaid/CHIP, also populate `federal_change` / `state_change` using the new variables.

## Why here vs. API

- Thin arithmetic on microsim sums — no Flask-specific state
- Analysis notebooks and ad-hoc Python scripts (not just the app) need this split
- policyengine-api#3481 becomes a pass-through wiring change once this lands

## Example

```python
from policyengine.tax_benefit_models.us import economic_impact_analysis

analysis = economic_impact_analysis(baseline_sim, reform_sim)
print(f"Federal cost: ${analysis.federal_budgetary_impact / 1e9:+.1f}B")
print(f"State cost:   ${analysis.state_budgetary_impact / 1e9:+.1f}B")
print(f"Total:        ${analysis.total_budgetary_impact / 1e9:+.1f}B")
```

## Related

- PolicyEngine/policyengine-us#8075 (scope)
- PolicyEngine/policyengine-us#8076 (Medicaid + CHIP attribution — merged)
- PolicyEngine/policyengine-api#3481 (API surface — depends on this)
- PolicyEngine/policyengine-app-v2#999 (UI — depends on API)
- PolicyEngine/analysis-notebooks#131 (example notebook — depends on this)

MaxGhenis:
Fixes #289

## Summary

Partitions US reform budgetary impact into federal, state, and **unattributed** shares:

- `BudgetaryImpact` (federal / state / unattributed, with a computed `total`) added to `PolicyReformAnalysis`
- `calculate_budgetary_impact(baseline_sim, reform_sim) -> BudgetaryImpact` helper
- Example output in `examples/us_budgetary_impact.py`

## Formulas

```
total        = Δ(household_tax) − Δ(household_benefits)
federal      = Δ(income_tax) + Δ(employee_payroll_tax) − Δ(federal_benefit_cost)
state        = Δ(state_income_tax)                      − Δ(state_benefit_cost)
unattributed = total − federal − state
```

Sign convention: positive means the government is better off (more revenue or lower spending). `total` is measured directly from the household aggregates, so it covers **every** tax and benefit the model computes. `federal` and `state` attribute only the pieces that are cleanly assignable to a level of government today; `unattributed` carries the residual — 100%-federal programs (SSI, SNAP, ...), 100%-state supplements, and any tax outside the three tax variables — until finer attribution exists. Because `total` is measured independently rather than summed from the attributed pieces, a reform to an unattributed program (e.g. SSI) surfaces in `total`/`unattributed` instead of silently reading as zero. `total` is a pydantic `computed_field`, so `federal + state + unattributed == total` by construction.

`federal_benefit_cost` / `state_benefit_cost` are the PE-US aggregates from PolicyEngine/policyengine-us#8076 that sum the federal/state shares of shared-funding programs (Medicaid FMAP, CHIP eFMAP). They are already present in the pinned `policyengine-us==1.752.2`, so **no pin bump is required**. They are not in the default US output, so `economic_impact_analysis` materializes them via `configure_budgetary_impact_variables` before running, mirroring how labor-supply and cliff-impact configure their extra outputs. As more programs gain attribution, the aggregates grow automatically via the parameter-driven `adds` list — no changes needed here.

## Why here vs. in policyengine-api

The arithmetic is thin sums over microsim outputs. Every consumer — API, notebooks, ad-hoc scripts — needs the same split; putting it behind the `.py` package means one implementation. The API follow-up (PolicyEngine/policyengine-api#3481) becomes a pass-through.

## Test plan

- [x] `pytest tests/test_budgetary_impact.py` — unit tests over the partition arithmetic (fully-attributed federal cut, shared-benefit federal/state split, unattributed SSI residual, mixed reform, zero reform, computed-total invariant)
- [x] Integration test runs `calculate_budgetary_impact` against a tiny hand-built US output simulation, resolving every referenced variable name against the installed policyengine-us model so a future rename fails loudly
- [x] `ruff format --check` + `ruff check` clean

## Related

- PolicyEngine/policyengine-us#8075 (scoping)
- PolicyEngine/policyengine-us#8076 (Medicaid + CHIP attribution, merged)
- PolicyEngine/policyengine-api#3481 (API surface, depends on this)
- PolicyEngine/policyengine-app-v2#999 (UI, depends on API)
- PolicyEngine/analysis-notebooks#131 (example notebook, depends on this)


vahid-ahmadi (changes requested):
The shape of this is right and the sign convention is consistent with the rest of the codebase, but I think there's a real bug in the federal-tax computation that the unit tests miss.

**Blocking: `payroll_tax` is not a variable in policyengine-us.**

Grepping the installed package: there is no `class payroll_tax` and no parameter-driven `payroll_tax` aggregate. Only `employee_payroll_tax` (TaxUnit) and `payroll_tax_gross_wages` (a base, not an aggregate) exist. This is the same rename that PR #327 just landed on the program-statistics side.

When `economic_impact_analysis` calls `calculate_budgetary_impact` → `_sum_change(..., "payroll_tax")` → `ChangeAggregate.run()` → `get_aggregate_variable(...)`, that helper (newly added in #327) will raise:

```
ValueError: ChangeAggregate.variable references missing variable 'payroll_tax' in policyengine_us version X.Y.Z. Did you mean: 'employee_payroll_tax'?
```

So this PR currently breaks `economic_impact_analysis` end-to-end on US. Suggest swapping to `employee_payroll_tax` to match #327.

The reason the test suite passes is that all four tests `patch("...analysis._sum_change", side_effect=...)` and feed in mocked deltas. The arithmetic is exercised but the variable names never touch policyengine-us. Worth one thin integration test that calls `_sum_change` against a real (tiny) simulation for each of the four variables — that's all it'd take to catch this and any future rename.

**Also worth resolving before merge**

- The PE-US pin bump for `federal_benefit_cost` / `state_benefit_cost` is called out as TODO in the description but isn't in the diff. If this lands before the pin bumps, the same `get_aggregate_variable` error fires from inside `calculate_budgetary_impact` for any consumer running `economic_impact_analysis`. Either bump the pin in this PR, or guard the budgetary-impact computation: try the aggregates; if the variables are missing in the installed PE-US, return `BudgetaryImpact(federal=federal_tax_change, state=state_tax_change, total=...)` with a logged warning, and make `budgetary_impact` `Optional[BudgetaryImpact]` on `PolicyReformAnalysis`. Either path keeps `economic_impact_analysis` working across the supported PE-US version range.

**Smaller comments**

- `state_income_tax` is at TaxUnit level (verified). `ChangeAggregate(... aggregate_type=SUM)` will weight by tax-unit weights, which is what you want — fine. PR #327 mentioned `household_state_income_tax` because it was specifically materializing per-household output for snapshots; that's a different use case from the national sum here. Keeping `state_income_tax` is correct.
- Sign convention check: medicaid rollback test expects `+90e9` federal when `federal_benefit_cost` drops by 90e9. `0 - (-90e9) = +90e9`. Consistent with "positive = government saves money" elsewhere. Good.
- `BudgetaryImpact.total` is recomputed at construction (`federal + state`) — would be slightly cleaner as a `@computed_field` so it can't be inconsistent if someone later constructs the model directly with mismatched fields. Optional.
- The example script will inherit whichever bug above hits first. Worth re-running it after the variable swap to confirm the printed numbers move in the expected direction.

**Verified correct**

- Sign convention matches the rest of the codebase.
- Arithmetic matches all four test cases under the assumed deltas.
- Scope notes are honest about which programs are omitted (100%-federal and 100%-state programs not in the shared-funding aggregates).
- `__init__.py` exports `BudgetaryImpact` and `calculate_budgetary_impact` cleanly.

Happy to re-review once `payroll_tax` → `employee_payroll_tax` and the PE-US pin / fallback question are resolved.

anth-volk:
@vahid-ahmadi does this relate to .py's stronger coupling of region and dataset together?

vahid-ahmadi:
@anth-volk Not really — both issues are variable-name coupling against the installed policyengine-us, not region/dataset coupling.

The `payroll_tax` failure is a straight rename: PE-US exposes `employee_payroll_tax`, not `payroll_tax` (same flavour as the program-statistics rename in #327). `ChangeAggregate.run()` resolves names against `tax_benefit_model_version.variables_by_name`, which is independent of which dataset is wired up. The pin question is similar — `federal_benefit_cost` / `state_benefit_cost` are new in policyengine-us#8076 and will fail on any installed PE-US below that, regardless of dataset.

Where region-coupling is adjacent to this PR: the variable list is silently US-specific (`income_tax`, `state_income_tax`, the federal/state benefit aggregates). `calculate_budgetary_impact` and `BudgetaryImpact` read as country-neutral but only make sense on a US `PolicyReformAnalysis`. Worth either a `_us` suffix on the helper, or wiring it through the same `tax_benefit_models/us/analysis.py` path that `economic_impact_analysis` already lives in. That's separate from the blocker though.

MaxGhenis:
Closing per stale-PR policy: conflicting with main since May. #289 stays open as the tracker; the economic-impact surface has moved since (see #258/#263), so a fresh cut is probably cheaper than a rebase.

MaxGhenis:
Reopening pending a proper per-PR review; specific disposition to follow.

MaxGhenis:
Revived after a merits review. Summary of changes since the last review:

- **Rebased onto `main`.** The only conflict was the `analysis.py` import block, resolved as the union of `main`'s cliff/labor-supply imports and this branch's `change_aggregate` imports; the `PolicyReformAnalysis` field interleaving and `__init__` exports auto-merged (stale household-impact refs from the pre-refactor base were dropped).
- **`payroll_tax` → `employee_payroll_tax`.** Blocker fixed: `payroll_tax` no longer exists in policyengine-us. `employee_payroll_tax` is the tax-unit-level employee-share tax, consistent with `household_net_income`. Code and docstrings updated.
- **No pin bump needed.** `federal_benefit_cost` / `state_benefit_cost` (policyengine-us#8076) are already present in the pinned `policyengine-us==1.752.2`, so `calculate_budgetary_impact` resolves on the pinned range without the guard/fallback path.
- **New total / partition semantics.** `total` is now measured directly as Δ`household_tax` − Δ`household_benefits` (every tax and benefit the model computes), not `federal + state`. `federal` and `state` attribute only the cleanly-assignable pieces (`income_tax` + `employee_payroll_tax` − `federal_benefit_cost`; `state_income_tax` − `state_benefit_cost`), and a new `unattributed` field carries the residual (`total − federal − state`). This fixes the silent-wrong-answer case: a reform to a 100%-federal program such as SSI or SNAP surfaces in `total`/`unattributed` instead of reading as ~0. `total` is a `@computed_field`, per your suggestion, so `federal + state + unattributed == total` by construction.
- **Integration test added.** A thin test runs `calculate_budgetary_impact` against a tiny hand-built US output simulation (mirroring `test_us_program_statistics.py`), so every referenced variable name (`household_tax`, `household_benefits`, `income_tax`, `employee_payroll_tax`, `state_income_tax`, `federal_benefit_cost`, `state_benefit_cost`) is resolved against the installed policyengine-us model and a future rename fails loudly.

Changelog fragment converted to `changelog.d/289.added.md`. `ruff format --check`, `ruff check`, and the budgetary-impact / program-statistics / parametric-reform / entity-utils tests pass locally.

MaxGhenis:
Second review round applied. An adversarial pass found that the revived total still mis-scored the shared-funding health programs themselves: Medicaid/CHIP/MSP are excluded from household_benefits under the default gov.simulation.include_health_benefits_in_net_income=false, so a Medicaid reform reported total ≈ 0 with a phantom offsetting unattributed residual — and the shared-benefit unit test encoded that impossible delta as ground truth. Now: total = Δhousehold_tax − Δhousehold_benefits − Δ(federal_benefit_cost + state_benefit_cost), with the cost subtraction skipped when the parameter folds health programs into household_benefits (double-count guard, tested). The shared-benefit test uses model-consistent deltas (household_benefits unmoved by a health reform), the mixed-reform and real-simulation tests are corrected to the same basis, and a guard-path test covers the parameter-true configuration. Also merged main (post-#458 shared program-statistics helpers; imports reconciled). Docstrings and the changelog fragment now state the actual basis including MSP.

PR #300, opened 2026-04-20
[v4.0.1] Fix contrib-CTC structural reforms in US microsim path

MaxGhenis:
## Summary

Fixes a silent bug in the US microsim path that made dict-reforms involving `gov.contrib.ctc.*` parameters crash at `.ensure()` with:
```
AttributeError: 'NoneType' object has no attribute 'entity'
```

Concretely: **Tara Watson's published CTC+EITC reform** ([policy 94589 on app.policyengine.org](https://www.policyengine.org/us/policy?reform=94589)) — 42 parameters with three `gov.contrib.ctc.*.in_effect = True` gates — was completely unable to run through `pe.us.economic_impact_analysis` on the v4.0 stack.

## Root cause

`PolicyEngineUSLatest._build_simulation_from_dataset` was instantiating entities against the **module-level** `policyengine_us.system`:

```python
builder.populations = system.instantiate_entities()   # module system, no user reform
# ...
microsim.build_from_populations(builder.populations)
```

But `Microsimulation.__init__` applies structural reforms (triggered by user-reform parameter gates like `gov.contrib.ctc.minimum_refundable.in_effect=True`) to **its own `tax_benefit_system`** — not the module one. Building populations against the module system then left reform-registered variables like `ctc_minimum_refundable_amount` invisible at calc time; `refundable_ctc` (which had been swapped to the reform version) would then crash trying to sum them.

## Fix

Pass `microsim.tax_benefit_system` to `_build_simulation_from_dataset` instead of the module-level `system`. One-line change in `src/policyengine/tax_benefit_models/us/model.py`.

## Evidence

On the v4.0 stack (pe-us 1.653.3 + us-data 1.73.0), running the full 42-param reform now produces:

| Metric | Baseline | Reform | Δ |
|---|---:|---:|---:|
| Federal income tax | $2,141.8B | $2,053.8B | **−$87.9B** |
| CTC | $138.2B | $236.3B | +$98.0B |
| Refundable CTC | $35.9B | $215.7B | +$179.8B |
| EITC | $61.4B | $38.9B | −$22.6B |

vs. pe-us 1.601.0 (March 2026): −$69.6B (the +$18B drift is upstream model evolution, not a pe.py regression).

## Tests

New `tests/test_us_microsim_structural_reforms.py` — builds a tiny in-memory dataset, applies the minimum-refundable reform via dict, runs the sim end-to-end. Before the fix: crashes. After: passes.

72/72 existing tests still pass.

## Version

Bumped to **4.0.1** (patch — bug fix, no API changes).


MaxGhenis:
### Subagent review complete — all blockers addressed

Two reviewers ran on this PR: code-simplifier and reproducibility-reviewer. Key findings, all addressed in commit e03dc8a:

**Reproducibility (blocking)**
- The original "module-level system stays pristine" invariant was brittle — would flip false if a future \`policyengine_us\` release ever shipped \`ctc_minimum_refundable_amount\` unconditionally. Replaced with a direct identity assertion via monkeypatch: the system passed into \`_build_simulation_from_dataset\` must \`is\` \`microsim.tax_benefit_system\`. That's the fix's actual contract; survives arbitrary upstream changes.

**Reproducibility (follow-ups)**
- Parameter-change-shows-up-in-output test now covers BOTH CTC AND EITC variables, catching partial-application regressions (not only no-op ones).
- Added one non-CTC gate (\`gov.contrib.streamlined_eitc\`) to the parametrised smoke test to prove the fix generalises beyond the CTC family.

**Code-simplifier**
- \`simple = lambda\` (with \`# noqa: E731\`) → \`def _simple\`.
- Fix's inline comment shortened from 6 lines to 4, kept the load-bearing "hide reform-registered variables" clause.

**UK latent bug**
- Reviewers flagged that UK's \`simulation_modifier\` pattern applies reforms AFTER \`Microsimulation.__init__\`, so any UK \`*.in_effect\`-gated structural reform would silently no-op. Filed as #302 for v4.0.2 / v4.1 — out of scope here.

**Verification**
- 7/7 tests pass with fix.
- Fix reverted → 3/7 tests fail with clean diagnostic errors (the invariant test now reports \`system is not microsim.tax_benefit_system\` instead of the raw \`NoneType\` traceback).

Ready for merge on CI green.

MaxGhenis:
Superseded by fresh rebase on main — see new PR.

Issue #326, opened 2026-04-30
Design a durable program-statistics mapping for changing country packages


anth-volk:
Follow-up to Vahid's JOSS PR: https://github.com/PolicyEngine/policyengine.py/pull/264

The immediate US breakage is tracked separately in #325. This issue is for the broader design problem: program-statistics analysis currently depends on hard-coded country program variable names and output assumptions, but country packages can rename, add, remove, or re-scope program variables over time.

The immediate fix can patch the current US mappings, validate them earlier, and improve the error message. That does not solve the underlying fragility. A durable design should make program-statistics mappings explicit, version-aware, and tested against the variables that are actually materialized into output datasets.

Potential directions:

- Maintain country-specific program-statistics metadata with explicit display names, concrete variable names, entities, tax/benefit polarity, and supported periods/package versions.
- Validate mappings before expensive simulations run, including both model-variable existence and output materialization.
- Prefer metadata or registry-driven program definitions where the country package exposes stable program concepts.
- Add drift-catching tests that load the current country model metadata and assert every configured program-stat variable can be aggregated from a small output fixture.
- Decide how to handle semantically ambiguous programs, especially cases like Medicare where the intended concept could be benefits, enrollment, payroll tax, or spending.

This should be treated as separate from the immediate repair PR, which is only intended to restore the currently broken US economic-impact path and make future failures easier to diagnose.

anth-volk:
Closing as completed. Program-statistics mappings are explicit, validated against model metadata, shared through common builders, and covered by tests.

PR #264, opened 2026-03-23
Add JOSS paper for submission

vahid-ahmadi:
## Summary
- Adds `paper.md` and `paper.bib` for submission to the [Journal of Open Source Software (JOSS)](https://joss.theoj.org/)
- Paper covers all 6 required JOSS sections: Summary, Statement of Need, State of the Field, Software Design, Research Impact Statement, and AI Usage Disclosure
- Cites adoption by 10 Downing Street, HM Treasury (Algorithmic Transparency Record), NIESR UK Living Standards Review 2025, INET Oxford (Seventh Carbon Budget), IEA reports, and House of Lords parliamentary debate
- Funded by the Nuffield Foundation

## What is JOSS?
[JOSS](https://joss.theoj.org/) is a peer-reviewed open-access journal for research software. It publishes short papers (750–1,750 words) describing open-source software with research impact. There are no submission fees. The review process happens on GitHub via [openjournals/joss-reviews](https://github.com/openjournals/joss-reviews). See the [submission guide](https://joss.readthedocs.io/en/latest/submitting.html) and [paper format](https://joss.readthedocs.io/en/latest/paper.html) for full details.

## JOSS requirements checklist
- [x] OSI-approved open-source license (AGPL-3.0)
- [x] Public repository with browsable source code
- [x] Issue tracker readable without registration
- [x] Public development history > 6 months
- [x] Demonstrated research impact (government, academic, policy citations)
- [x] Good open-source practices (CI, tests, docs, PRs, tagged releases)
- [x] Iterative development over time
- [x] paper.md with all 6 required sections
- [x] paper.bib with BibTeX references
- [x] AI Usage Disclosure
- [x] Funding acknowledgement (Nuffield Foundation)
- [x] ORCID IDs for all authors

## Review suggestions status
From the [self-review comment](https://github.com/PolicyEngine/policyengine.py/pull/264#issuecomment-4170019413):

- [x] **Add architecture diagram** — architecture.png added with proper Figure 1 reference in text
- [x] **Fix placeholder ORCIDs** — All 4 authors have real ORCIDs
- [x] **Add a minimal code example** — No code snippet yet (QuantEcon-style)
- [x] **Add comparison table in State of the Field** — Feature table vs TAXSIM/UKMOD/OpenFisca
- [x] **Add reproducibility framing to Statement of Need** — Added: "Because existing proprietary models cannot be independently verified..."
[...]

PR #327, opened 2026-04-30
Fix US program statistics variable mappings

### LINKED ISSUE #325: US economic_impact_analysis uses fragile hard-coded program variable names
Follow-up from the instrumented US budgetary-impact run and PR discussion: https://github.com/PolicyEngine/policyengine.py/pull/264#issuecomment-4355958830

`economic_impact_analysis` currently builds US program statistics from a hard-coded list of program variable names. The run completed the expensive baseline and reform simulations and produced quick aggregate outputs, but then failed in `ProgramStatistics -> Aggregate.run()` with `StopIteration` while looking up a configured variable in `simulation.tax_benefit_model_version.variables`.

Likely immediate mismatch: the US program list includes `payroll_tax`, but the US model/output path appears to expose/materialize `employee_payroll_tax` instead. Other entries may also be fragile, for example `medicare` may not correspond to a plain aggregate variable in the US package.

This pattern is especially fragile for the US package because variable names and output variables can change quickly. The analysis layer should probably either:

- validate the configured program list up front and raise a clear error listing missing variables,
- derive supported program statistics from model metadata/output metadata, or
- maintain an explicit country-versioned mapping that is tested against current US outputs.

The important distinction from the run: the final exception is an analysis-layer lookup bug, while the earlier VSCode crash was more consistent with resource pressure during the simulation phase.

anth-volk:
Fixes #325

This draft PR addresses only the immediate causes of the US `economic_impact_analysis` program-statistics breakage observed while following up on Vahid's JOSS PR (#264). It is not intended to solve the broader design problem of country package program variables changing over time; that follow-up is tracked in #326.

Changes:

- Keep `ProgramStatistics.program_name` as the direct model/output variable to aggregate; no separate alias field is introduced.
- Replace the broken US program-statistics mappings with currently valid `policyengine-us` variables:
  - `payroll_tax` -> `employee_payroll_tax`
  - `medicare` -> `medicare_cost`
  - `state_income_tax` remains `state_income_tax`
- Note: `payroll_tax` was configured as a person-level program, while `employee_payroll_tax` is a tax-unit-level variable. This means the payroll-tax program-statistics row now reports tax-unit recipient/winner/loser counts, using tax-unit weights, rather than person counts.
- Add `medicare_cost` and `state_income_tax` to the US output variables materialized by `.py`.
- Update US household snapshots for the newly exposed `medicare_cost` and `state_income_tax` output keys.
- Validate US program-statistics configuration before expensive simulations run.
- Add shared error helpers for constructing typed errors and conditional error detail lines, then use them in the US program-statistics validation error path.
- Replace bare aggregate `next(...)` lookup failures with descriptive `ValueError`s.
- Add mocked US program-statistics, aggregate error, and error-helper tests that would have caught this issue.

Verification:

- `make format`
- `uv run --python 3.13 --extra dev python -m pytest tests/test_household_calculator_snapshot.py::test_us_household_snapshot tests/test_errors.py tests/test_aggregate.py tests/test_change_aggregate.py tests/test_us_program_statistics.py`
- `uv run --python 3.13 --extra dev pytest tests/test_change_aggregate.py tests/test_us_program_statistics.py`
- `uv run --python 3.13 --extra dev ruff check tests/test_us_program_statistics.py tests/test_change_aggregate.py`
- `git diff --check`


vahid-ahmadi:
## Review

Core logic is correct — variable name swaps, validation, and `StopIteration` → `ValueError` upgrade all check out. Two things to fix before merge.

### Blocking

1. **CI failures are caused by this PR.** Adding `medicare_cost` to `entity_variables["person"]` makes the household calculator materialize it, and 4 snapshots in `test_household_calculator_snapshot.py` fail with `new key: person[0].medicare_cost=14500.0`. Need to regenerate the snapshot files.

2. **Negative test can silently pass on the wrong exception.** `test_us_program_statistics_config_fails_before_simulation_run` uses `try/except` + manual `raise AssertionError`. Use `pytest.raises(ValueError, match="US program statistics config is invalid")` instead.

### Worth addressing

3. **`ChangeAggregate` has no invalid-variable regression test.** `test_aggregate.py` got one for `Aggregate`, but `ChangeAggregate` had the same bare `next(...)` pattern and isn't covered.

4. **Entity change not flagged in the PR description.** `payroll_tax` was `entity="person"`; `employee_payroll_tax` is `entity="tax_unit"`. Correct, but it changes recipient counts in the program-statistics table — worth noting.

### Verified correct

- All three variable replacements match what `policyengine-us` actually exposes (`employee_payroll_tax`, `medicare_cost`, `household_state_income_tax`).
- Validator runs cheap, before `ensure()` on either simulation.
- Test arithmetic: `employee_payroll_tax` 1600, `medicare_cost` 300, `state_income_tax` 900 — all correct under the test weights.
- Stays in scope for #325; no drift into #326.

### Minor suggestions

- `context=` strings like `"Aggregate.variable"` are internal field names; user-facing wording reads better.
- Missing return annotations on the three new helpers (`-> None`, `-> Variable`).
- Validator dedupes errors across baseline/reform — asymmetric configs harder to triage.

vahid-ahmadi (commented):
Follow-up review on the new commits (`fe78636` … `0c39e5a`).

**Still worth addressing**

1. **`create_error` is a thin wrapper that adds nothing.** Body is `return error_type(message)`, and the test (`test_create_error_returns_requested_error_type`) just re-asserts Python's exception constructor behavior. Both `raise create_error(ValueError, msg)` and `raise ValueError(msg)` produce identical results. The indirection makes the call site longer without helping callers. Suggest removing `create_error` (and its test) and inlining `raise ValueError(...)`. `format_conditional_error_detail` should stay — it has real logic (sort + dedupe + None-on-empty).

2. **Asymmetric baseline/reform configs still merge errors.** `_validate_program_statistics_config` puts missing variables and missing outputs from both simulations into the same sets, so a config where baseline lacks `medicare_cost` and reform lacks `social_security` produces one message that doesn't say which simulation each is missing from. Was minor before, still minor — just noting it's unchanged.

3. **Inconsistent return annotations on the new aggregate helpers.** `require_output_column` has `-> None`; `get_aggregate_variable` and `get_output_entity_data` don't. Easy fix — return `Variable` and the entity data type respectively, or at least `Any`.

**New observation**

The `match=\"Aggregate.variable\"` / `match=\"ChangeAggregate.filter_variable\"` test assertions now lock the internal `context=` strings into the test contract. If those labels ever get rewritten to be more user-facing (which was a prior minor suggestion), the tests need to update too. Not a blocker — just flagging the new coupling.

Looks mergeable once `create_error` is removed (or kept with a note explaining what it gives over `ValueError(msg)`).

vahid-ahmadi (commented):
Follow-up after `ed1ab89`.

**Addressed**

- `create_error` removed everywhere (utils/errors.py, utils/__init__.py re-export, analysis.py import, the tautological test). Inline `raise ValueError(...)` used in the validator. 
- Return annotations added on the new aggregate helpers — `get_aggregate_variable(...) -> Variable`, `get_output_entity_data(...) -> Any`, plus `data: Any` on `require_output_column`.
- Test contract loosened nicely: instead of `match=\"Aggregate.variable\"` (which would have locked the internal `context=` label into the test), the tests now assert on the user-facing fragments `\"references missing variable\"` plus the variable name. That defuses the coupling I flagged last round and leaves room to reword `context=` later without churn.

The asymmetric baseline/reform-config message is still merged, but that was minor and explicitly noted as such — fine to leave.

LGTM. Nothing else from me; ready to merge.

vahid-ahmadi (approved):

PR #334, opened 2026-05-01
Derive program-statistics entity from model metadata (refs #326)

vahid-ahmadi:
Refs #326. **Draft** — see scope notes and overlap with #327 below.

## Summary
Replaces the hard-coded `programs = {...}` dict in `economic_impact_analysis` (US) with a structured `ProgramSpec` list and a `resolve_program_specs(specs, model_version)` helper that:

- **Validates every variable name** against the model up front, before any expensive simulation work.
- **Derives \`entity\` from the variable's own metadata** — the entity string is no longer duplicated next to the variable name, so it cannot silently drift when `policyengine-us` moves a variable between entities.
- **Collects every unknown variable into a single \`ValueError\`** with `difflib` suggestions, instead of failing one-at-a-time deep inside an `Aggregate.run()` `next(...)` lookup.

This is a concrete first step toward the durable program-statistics mapping tracked in #326. Deriving the *list itself* from model metadata (e.g. scanning for variables tagged as program aggregates) is left for a follow-up.

## Files
- `src/policyengine/tax_benefit_models/us/programs.py` (new) — `ProgramSpec`, `ResolvedProgram`, `US_PROGRAM_SPECS`, `resolve_program_specs()`.
- `src/policyengine/tax_benefit_models/us/analysis.py` — uses the new helper.
- `tests/test_program_specs.py` (new) — 4 unit tests covering entity derivation, multi-error collection, fuzzy suggestions, and de-duplication.
- `changelog.d/program-specs-metadata.changed.md`.

## Overlap with #327
This PR overlaps with @anth-volk's #327 (the immediate fix for the `StopIteration` crash in #325). The variable name corrections from #327 (`payroll_tax` → `employee_payroll_tax`, `medicare` → `medicare_cost`, `state_income_tax` → `household_state_income_tax`) are applied here too so the new validation passes against the current US model.

Whichever lands first, the other needs a small rebase. Marked draft for that reason.

## Test plan
- [ ] `pytest tests/test_program_specs.py`
- [ ] `pytest tests/test_aggregate.py tests/test_change_aggregate.py` still passes.
- [ ] Manually run `examples/us_budgetary_impact.py` and confirm `economic_impact_analysis` produces the program-by-program table without the original `StopIteration`.


vahid-ahmadi:
Superseded by #342, which lands the same entity-derivation idea on top of merged #327 (smaller diff, keeps the materialized-outputs check). Closing this draft.

PR #338, opened 2026-05-03
Add CRFB TOB scenario contract

MaxGhenis:
## Summary
- add a named CRFB Post-OBBBA long-run TOB scenario contract
- expose constants for the target source, scenario id, target hash, and trustees-2025 core threshold law mode
- add metadata validation helpers that reject raw current-law substitution and hash mismatches

## Validation
- ruff check src/policyengine/scenarios tests/test_crfb_scenario_contract.py
- ruff format --check src/policyengine/scenarios tests/test_crfb_scenario_contract.py
- PYTHONPATH=src /Users/maxghenis/PolicyEngine/policyengine.py/.venv/bin/python -m pytest tests/test_crfb_scenario_contract.py

## Notes
- This package carries only the portable scenario contract. It does not try to load policyengine-us-data or imply the long-term model/artifacts are installed.
- The first pytest attempt with PYTHONPATH=. failed because this repo uses a src layout; rerun with PYTHONPATH=src passed.

### DIFF
--- a/changelog.d/crfb-tob-scenario.added.md
+++ b/changelog.d/crfb-tob-scenario.added.md
@@ -1,0 +1,1 @@
+Added a named CRFB Post-OBBBA long-run TOB scenario contract for validating the target source, hash, and Trustees threshold mode.
--- a/src/policyengine/scenarios/__init__.py
+++ b/src/policyengine/scenarios/__init__.py
@@ -1,0 +1,19 @@
+from policyengine.scenarios.crfb import (
+    CRFB_POST_OBBBA_TOB_SCENARIO_ID,
+    CRFB_POST_OBBBA_TOB_TARGET_SHA256,
+    CRFB_POST_OBBBA_TOB_TARGET_SOURCE,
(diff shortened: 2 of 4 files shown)


MaxGhenis:
Reviewed 2026-07-08 as part of the open-PR sweep — disposition: stays open as an active draft. It's self-contained (new policyengine.scenarios namespace, nothing on main imports it), all checks green. Deliberately NOT merging while the CRFB engagement is held, for two reasons: (1) merging publishes a new public namespace (7 exported symbols) to PyPI as a compatibility surface; (2) the contract hardcodes a provisional calibration target (oact_2025_08_05_provisional + its sha256) that may change before the engagement finalizes. Pre-merge checklist when CRFB resumes: reconcile target_source/target_sha256 with the finalized OACT artifact, and make a deliberate call on whether policyengine.scenarios is the right public home vs keeping the contract engagement-side. Non-blocking nit: validate_metadata compares metadata['name'] against the contract's target_source field — confusing key/field naming.

PR #340, opened 2026-05-04
Reject unsupported monthly household inputs

### LINKED ISSUE #339: Household calculators should reject monthly periods and periodized inputs until supported
## Problem

The household calculators expose a `year` argument, but because Python does not enforce the `int` type hint at runtime, callers can pass monthly period strings such as `year="2026-01"`. In the US path, this reached the underlying country simulation and returned values, but the results were not semantically reliable.

The household wrappers periodize every household input under `str(year)`, so a monthly string causes annual inputs such as `employment_income` to be stamped onto a monthly period. In local investigation, this produced distorted output: annual inputs disappeared or changed, while monthly variables such as SNAP still returned monthly-looking values.

The helpers also do not support periodized household input values such as:

```python
people=[{"employment_income": {"2026-01": 1000}}]
```

Those values are currently ambiguous because the helpers wrap input values themselves as `{year: value}`.

## Expected behavior

Until monthly household calculations are explicitly designed and tested, household helpers should fail early when callers provide:

- monthly/non-annual `year` values such as `"2026-01"`
- periodized household input values such as `{"2026-01": 1000}`

Reform effective-date dictionaries should remain valid, since those are separate from household input values and are already part of the supported reform API.

## Scope

This issue is only for an immediate guardrail. It is not intended to implement full monthly household support or redesign the flattened HDF5 dataset format.

anth-volk:
Fixes #339

## Summary

This PR adds an immediate shared guardrail to the household calculators while monthly household support remains undesigned/unsupported.

It now fails early when callers pass:

- a monthly or otherwise non-annual `year` value such as `"2026-01"`
- periodized household input values such as `{"2026-01": 1000}`

The validation is shared across the US and UK household calculators and uses country-neutral error messages. It preserves supported reform effective-date dictionaries, e.g. `reform={...: {"2026-01-01": value}}`.

## Notes

This is intentionally narrow. It does not implement full monthly household support and does not change the annual flattened HDF5 dataset structure.

## Tests

- `python -m py_compile src/policyengine/tax_benefit_models/common/household.py src/policyengine/tax_benefit_models/us/household.py src/policyengine/tax_benefit_models/uk/household.py tests/test_household_impact.py`
- `uv run pytest tests/test_household_impact.py -q`
- `uv run pytest tests/test_household_impact.py -q -k "monthly_year_period or periodized_person_input or periodized_group_input"`
- `uv run ruff check src/policyengine/tax_benefit_models/common/household.py src/policyengine/tax_benefit_models/common/__init__.py src/policyengine/tax_benefit_models/us/household.py src/policyengine/tax_benefit_models/uk/household.py tests/test_household_impact.py`
- `uv run ruff format --check src/policyengine/tax_benefit_models/common/household.py src/policyengine/tax_benefit_models/common/__init__.py src/policyengine/tax_benefit_models/us/household.py src/policyengine/tax_benefit_models/uk/household.py tests/test_household_impact.py`

### DIFF
--- a/src/policyengine/tax_benefit_models/common/__init__.py
+++ b/src/policyengine/tax_benefit_models/common/__init__.py
@@ -6,6 +6,9 @@ their public ``calculate_household`` / ``analyze_reform`` entry points.
 """
 
 from .extra_variables import dispatch_extra_variables as dispatch_extra_variables
+from .household import (
+    validate_annual_household_inputs as validate_annual_household_inputs,
+)
 from .model_version import (
     MicrosimulationModelVersion as MicrosimulationModelVersion,
 )
--- a/src/policyengine/tax_benefit_models/common/household.py
+++ b/src/policyengine/tax_benefit_models/common/household.py
@@ -1,0 +1,59 @@
+from __future__ import annotations
+
+from collections.abc import Mapping, Sequence
+from typing import Any
+
+
+def validate_annual_household_inputs(
+    *,
+    year: Any,
+    entities: Mapping[str, Sequence[Mapping[str, Any]]],
+) -> int:
+    """Validate annual-only household calculator inputs."""
+    validated_year = _validate_annual_year(year)
+    _validate_unperiodized_values(entities)
+    return validated_year
+
+
+def _validate_annual_year(year: Any) -> int:
+    if isinstance(year, bool):
+        raise _annual_period_error()
+    if isinstance(year, int):
+        return year
+    if isinstance(year, str) and year.isdecimal() and len(year) == 4:
+        return int(year)
+    raise _annual_period_error()
+
+
+def _annual_period_error() -> ValueError:
+    return ValueError(
+        "Household calculations require a calendar year as an integer, "
+        "for example year=2026. Monthly periods are not supported by "
+        "calculate_household."
+    )
+
+
+def _validate_unperiodized_values(
+    entities: Mapping[str, Sequence[Mapping[str, Any]]],
+) -> None:
+    for entity, records in entities.items():
+        for index, record in enumerate(records):
+            for variable, value in record.items():
+                if variable != "id" and isinstance(value, Mapping):
+                    raise ValueError(
+                        "Periodized household inputs are not supported by "
+                        "calculate_household. Pass annual scalar input values "
+                        f"only; received a periodized value for "
+                        f"{_input_location(entity, index, len(records), variable)}."
+                    )
+
+
+def _input_location(
+    entity: str,
+    index: int,
+    record_count: int,
+    variable: str,
+) -> str:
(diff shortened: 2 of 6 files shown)


vahid-ahmadi (commented):
Well-scoped guardrail that does exactly what the issue asked for. The shared `common/household.py` is the right home, validation runs before the country-package import, and the tests cover both UK and US for the year check and the periodized-value check on person and group entities.

**Minor comments**

1. **UK docstring not updated.** `src/policyengine/tax_benefit_models/uk/household.py` still says only "unknown or mis-placed variable names, or unknown reform parameter paths" under `Raises:`. The US docstring was updated; worth mirroring on UK for parity.
2. **Behavior change for `year` as string.** Before this PR, `year=\"2026\"` flowed through as a string. After, it's normalized to `int(2026)`. Almost certainly an improvement but not called out in the PR description — worth a line.
3. **`_validate_unperiodized_values` only flags `Mapping` values.** A list like `employment_income=[1000, 2000]` won't be caught. That's outside the issue's scope (which named periodized dicts specifically), so fine — just flagging in case the guardrail expands later.
4. **Soft coupling in `_input_location`.** The `record_count == 1 and entity != \"people\"` branch works because callers always pass group entities as `[value]` and the `people` key for the multi-record list. Acceptable since the function is module-private and both call sites are in this PR.
5. **Optional: include the offending year in the error.** `_annual_period_error()` is constant; adding `f\"received year={year!r}\"` would make the failure slightly more debuggable.

vahid-ahmadi (commented):
**One observation on the new approach**

Instead of normalizing `"2026"` → `2026` once at the boundary, this commit widens the type to `Union[int, str]` and threads it through `_build_situation`, `compile_reform`, `_reform_dict_to_parameter_values`, `_compile_reform_to`, `compile_reform_to_policy`, `compile_reform_to_dynamic`. Functionally fine — `f"{year}-01-01"` and `str(year)` work for both — but it does mean every public reform-compilation entry point now exposes the wider type to its own callers.

Either approach is defensible; the verbatim path is more honest about caller intent, the normalize-early path keeps downstream signatures narrower. If you ever want to revisit, the smaller-blast-radius option is to normalize in `validate_annual_household_inputs` and keep `compile_reform`'s `year` as `int`. Not blocking.

**Tiny nit**

`AnnualYear = Union[int, str]` is defined in `common/household.py` but `compile_reform` declares the same union inline. If the alias is meant to be the canonical name for "a year a household calculator accepts", export it from `common/__init__.py` and reuse it in reform.py; otherwise drop the alias and inline both. Either way is fine — just inconsistent right now.

LGTM to merge.

anth-volk:
Thanks for review @vahid-ahmadi. I had actually intended to address your first item in my changes. I will fix that, push, then merge.

PR #342, opened 2026-05-07
Derive US program-statistics entity from variable metadata

vahid-ahmadi:
Refs #326. Smaller follow-up on top of merged #327. Replaces #334.

## Summary
- Drop the duplicated `entity` field from `US_PROGRAMS`. Entity is now derived from `model_version.get_variable(name).entity` at runtime, so the program list cannot silently drift when `policyengine-us` moves a variable between entities.
- Keeps the "missing outputs" check that #327 added in `_validate_program_statistics_config` (variable must be both defined in the model **and** materialized in the simulation outputs).
- Adds `test_us_programs_entities_match_model_metadata` so a typo or rename in `US_PROGRAMS` is caught at test time, not at simulation time.

## Why this instead of #334
#334 was authored before #327 landed and (a) is severely behind `main`, (b) drops the materialized-outputs check from #327, and (c) introduced a `programs.py` module / `ProgramSpec` dataclasses that aren't load-bearing for the entity-derivation goal. This PR keeps only the durable improvement and stays minimal.

## Test plan
- [ ] `pytest tests/test_us_program_statistics.py` (covers entity derivation in `economic_impact_analysis` and the new metadata-match test).
- [ ] `pytest tests/test_aggregate.py tests/test_change_aggregate.py` still passes.

### DIFF
--- a/changelog.d/derive-us-program-entities.changed.md
+++ b/changelog.d/derive-us-program-entities.changed.md
@@ -1,0 +1,1 @@
+Derive US program-statistics entity from variable metadata instead of duplicating it in the program list.
--- a/src/policyengine/tax_benefit_models/us/analysis.py
+++ b/src/policyengine/tax_benefit_models/us/analysis.py
@@ -28,18 +28,23 @@ from policyengine.outputs.poverty import (
 )
 from policyengine.utils.errors import format_conditional_error_detail
 
-US_PROGRAMS = {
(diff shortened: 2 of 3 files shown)


anth-volk (commented):
Not blocking, but I'd prefer `is_tax` remain a key. It's unclear to me what US_PROGRAMS actually represents short of reading the comments.

anth-volk (approved):

PR #344, opened 2026-05-07
Update UK release bundle to data 1.53.0

### LINKED ISSUE #343: Refresh UK bundled release contract to a TRO-compatible data release
The bundled UK release contract currently points at `policyengine-uk==2.88.0` and `policyengine-uk-data==1.40.4`. That data release does not publish the release metadata or certified artifact SHA needed by the TRACE TRO generation path, which caused the latest `policyengine.py` versioning deployment to fail while regenerating bundled TROs.

This issue tracks a stopgap update that:

- Moves the bundled UK pair to `policyengine-uk==2.88.6` and `policyengine-uk-data==1.53.0`.
- Records the certified UK dataset artifact SHA and data build fingerprint in the bundled UK manifest.
- Regenerates the bundled UK TRACE TRO sidecar.
- Updates the refresh helper so future bundle refreshes use the data release manifest metadata when one is available, instead of leaving certification fields stale.

This is not the full cross-repository bundle system. It is a targeted fix so the current `.py` release/versioning path can publish a UK contract whose TRO generation has the hashes it requires.

anth-volk:
Fixes #343

## Summary

This PR updates the bundled UK release contract to a TRO-compatible release pair:

- `policyengine-uk==2.88.6`
- `policyengine-uk-data==1.53.0`

It records the certified `enhanced_frs_2023_24` artifact SHA and data build fingerprint in the UK manifest, updates the package pins and lockfile, refreshes tests, and adds the generated bundled TRACE TRO sidecars.

It also fixes `refresh_release_bundle` so future refreshes use `release_manifest.json` from the target data release when available. That prevents stale certification metadata, including `built_with_model_version`, from being carried forward after data/model bumps.

## Current release-version behavior

The generated manifest/TRO files in this PR still show the current checked-in `policyengine.py` version, `4.3.1`. That is an inconsistency in the current `.py` release workflow, not the final published identity for this change: after this merges, the `Versioning` workflow runs Towncrier, bumps `pyproject.toml`, syncs `policyengine_version` and `bundle_id` in the bundled manifests, regenerates the TRO sidecars, and commits those release artifacts before the package is published.

The forthcoming `policyengine-bundles` system should avoid this inconsistency by making the bundle version/digest the canonical certified artifact first. A `.py` release would then point to an already-published bundle version/digest rather than generating provisional pre-release TROs that are later rewritten by `.py` versioning.

## UK validation requested

@vahid-ahmadi please confirm whether this `policyengine-uk==2.88.6` / `policyengine-uk-data==1.53.0` combination is valid and should be user-facing as the bundled UK runtime pair.

## Verification

- `uv run --extra dev pytest tests/test_bundle_refresh.py`
- `uv run --extra dev pytest tests/test_release_manifests.py tests/test_trace_tro.py`
- `ruff format --check .`
- `ruff check .`

## Notes

This is a stopgap update for the current `.py` release path. It does not replace the broader cross-repository bundle/orchestrator plan.

### DIFF
--- a/changelog.d/update-uk-release-bundle-1-53-0.changed.md
+++ b/changelog.d/update-uk-release-bundle-1-53-0.changed.md
@@ -1,0 +1,1 @@
+Updated the bundled UK release contract to policyengine-uk 2.88.6 and policyengine-uk-data 1.53.0.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,7 @@ graph = [
 ]
 uk = [
     "policyengine_core>=3.25.0",
-    "policyengine-uk==2.88.0",
+    "policyengine-uk==2.88.6",
 ]
 us = [
     "policyengine_core>=3.25.0",
@@ -60,7 +60,7 @@ dev = [
     "pytest-asyncio>=0.26.0",
     "ruff>=0.9.0",
     "policyengine_core>=3.25.0",
-    "policyengine-uk==2.88.0",
+    "policyengine-uk==2.88.6",
     "policyengine-us==1.687.0",
     "towncrier>=24.8.0",
     "mypy>=1.11.0",
--- a/src/policyengine/data/release_manifests/uk.json
+++ b/src/policyengine/data/release_manifests/uk.json
@@ -5,29 +5,31 @@
   "policyengine_version": "4.3.1",
   "model_package": {
     "name": "policyengine-uk",
-    "version": "2.88.0",
-    "sha256": "46a3ba443b43ec810c5efaccd4645edb63c8dc90ef5acf9b0cdf5ace86b9334d",
-    "wheel_url": "https://files.pythonhosted.org/packages/23/7e/8a2a42eac1da63730a865964aa17e7fd4420ce4db4c80001c1b5ca6011e8/policyengine_uk-2.88.0-py3-none-any.whl"
+    "version": "2.88.6",
+    "sha256": "d89c86c4835629115d9c20005dc818462798ac1940616f788db37e2b87472e33",
+    "wheel_url": "https://files.pythonhosted.org/packages/b8/38/3cf1f123e0ea25f16e3e207d46ba7ca86e6c251e3171dd7d22b18723c9a5/policyengine_uk-2.88.6-py3-none-any.whl"
   },
   "data_package": {
     "name": "policyengine-uk-data",
-    "version": "1.40.4",
+    "version": "1.53.0",
     "repo_id": "policyengine/policyengine-uk-data-private"
   },
   "certified_data_artifact": {
     "data_package": {
       "name": "policyengine-uk-data",
-      "version": "1.40.4"
+      "version": "1.53.0"
     },
-    "build_id": "policyengine-uk-data-1.40.4",
+    "build_id": "policyengine-uk-data-1.53.0",
     "dataset": "enhanced_frs_2023_24",
-    "uri": "hf://policyengine/policyengine-uk-data-private/enhanced_frs_2023_24.h5@1.40.4"
+    "uri": "hf://policyengine/policyengine-uk-data-private/enhanced_frs_2023_24.h5@1.53.0",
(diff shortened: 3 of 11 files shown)


anth-volk:
@vahid-ahmadi requested your review here. The main thing to confirm is whether `policyengine-uk==2.88.6` with `policyengine-uk-data==1.53.0` is a valid UK runtime pair and should be user-facing as the bundled UK default.

anth-volk:
Blocked on https://github.com/PolicyEngine/policyengine-uk/issues/1666

MaxGhenis:
Update after merging/releasing the UK Marriage Allowance fix: policyengine-uk==2.88.14 is now on PyPI and fixes the raw child Marriage Allowance scenario, but it is not enough to safely update this PR.\n\nI tested bumping this branch locally from 2.88.6 to 2.88.14. Import fails before the snapshot test with:\n\nValueError: Runtime data build fingerprint does not match the bundled data certification.\n\nThat is the certification guard doing the right thing. The new UK package reports data_build_fingerprint sha256:dff5c5bb976ce254fa965ecfce6a0d84859fe1629a714ae28a79b3075522a0ae, while this PR bundle currently certifies data 1.53.0 with sha256:535d2fd64f9e2b2aa9991e5a5d65be25bdd126f18858682533789f5c7e467782.\n\nSo the next safe step is data-side: publish/finalize a UK 1.53.0 release manifest that explicitly certifies compatibility with policyengine-uk==2.88.14, or regenerate the UK data release with 2.88.14. PR PolicyEngine/policyengine-uk-data#397 is now merged, so the manifest finalization path has the strict bundle fields needed for that. I did not push the local 2.88.14 bump because doing so would require weakening or falsifying the bundled certification.

vahid-ahmadi:
**Edited:** my earlier code-side push (`97c0bb4`) has been reverted (`0b4b7be`) and the analysis below is updated. After diffing against `main`, both halves of this PR have been independently overtaken on `main`, so the `bundle.py` fix I contributed is no longer load-bearing.

## Where `main` stands today

| Field | This PR | `main` |
| --- | --- | --- |
| `policyengine-uk` | `2.88.6` | **`2.88.20`** |
| `policyengine-uk-data` | `1.53.0` | **`1.55.10`** |
| `policyengine-us` | `1.687.0` | **`1.700.0`** |
| `policyengine_core` | `>=3.25.0` | **`==3.26.1`** |

`main`'s UK pair is well past the `2.88.14` cutoff that fixes `policyengine-uk#1666` (children receiving Marriage Allowance / income tax), so @MaxGhenis's original blocker is now resolved at HEAD.

## `bundle.py` direction has also landed on `main` separately

`main` has implemented the "use the data release manifest as the source of build provenance" idea on its own track (`_fetch_data_release_manifest` returning `_DataReleaseManifestFetch` with `x-repo-commit` header capture and a `main`-revision fallback for repos that don't tag every data version). The drift is 461 lines on `bundle.py` alone, and the two implementations are alternative designs of the same feature rather than something that can be mechanically merged.

## Validation answer to your original ask

`policyengine-uk==2.88.6` + `policyengine-uk-data==1.53.0` is **not** the pair to ship — CI on this branch still fails `test_uk_household_snapshot[uk_couple_two_kids]` with the children-income-tax regression. But that's now moot: `main` is at `2.88.20` / `1.55.10`, which is the right user-facing pair.

## Status

- 🔁 My code contribution to this PR has been reverted (`0b4b7be`) — `main`'s `bundle.py` already covers the same ground.
- ❌ The pin bump in this PR is superseded by `main`.
- ⚠️ Merge conflicts remain (pyproject.toml, uk.json, both `.trace.tro.jsonld` files, `bundle.py`, four test files, `uv.lock`); resolving them mechanically would produce a no-op since every conflict resolves to "take `main`".

Happy to revisit if there's a unique idea here you still want to land — otherwise this PR is effectively a no-op vs. `main`.

vahid-ahmadi (changes requested):
pls see my comment

anth-volk:
Closing as stale

PR #358, opened 2026-05-13
Add UK wealth decile impact outputs

anth-volk:
Fixes #353
Fixes #354

## Summary

- Adds `decile_variable` support to `calculate_decile_impacts(...)`, allowing callers to measure changes in one variable while grouping by a precomputed decile variable.
- Adds UK `wealth_decile_impacts` and `intra_wealth_decile_impacts` to `economic_impact_analysis(...)`.
- Documents the new helper argument, UK-only analysis fields, and the additive `decile_variable` dataframe column.

## Testing

- `uv run --extra dev python -m pytest tests/test_intra_decile_impact.py tests/test_uk_analysis.py -q`
- `uv run --extra dev ruff check src/policyengine/outputs/decile_impact.py src/policyengine/tax_benefit_models/uk/analysis.py tests/test_intra_decile_impact.py tests/test_uk_analysis.py`
- `uv run --extra dev ruff format --check src/policyengine/outputs/decile_impact.py src/policyengine/tax_benefit_models/uk/analysis.py tests/test_intra_decile_impact.py tests/test_uk_analysis.py`

## Notes

- This intentionally does not update `policyengine-api-v2`; API response mapping for `wealth_decile` and `intra_wealth_decile` will be handled separately.
- `calculate_decile_impacts(...).dataframe` now includes an additive `decile_variable` column so consumers can distinguish income-derived deciles from precomputed grouping variables.

PR #359, opened 2026-05-13
Add UK detailed budget program statistics

### LINKED ISSUE #352: Determine whether pre-1 detailed_budget maps to ProgramStatistics
## Context

`policyengine-api-v2` needs to preserve the legacy single-year macro output shape when it eventually moves back to `policyengine` 4.x. The pre-1 `Simulation.calculate_economy_comparison()` result included `detailed_budget`, while the 4.x `economic_impact_analysis()` path exposes `program_statistics`.

Before the simulation API can safely adapt 4.x outputs back into the legacy response contract, we need to determine whether `detailed_budget` can be reconstructed from `ProgramStatistics` without semantic drift.

## Questions to answer

- Does every pre-1 `detailed_budget` entry have a corresponding `ProgramStatistics` row for US and UK?
- Are signs, baseline/reform totals, changes, and recipient/count fields equivalent?
- Are labels/groupings in the old `detailed_budget` public contract recoverable from 4.x metadata?
- Are any old `detailed_budget` fields intentionally unsupported in 4.x?

## Acceptance criteria

- Document the mapping from pre-1 `detailed_budget` to 4.x `ProgramStatistics` for US and UK.
- Identify any unmappable fields or sign convention differences.
- Add tests or fixtures that compare representative pre-1 and 4.x outputs if feasible.
- Recommend whether the sim API should reconstruct `detailed_budget` from `ProgramStatistics` or expose a new dedicated `.py` output helper.

anth-volk:
Fixes #352

## Summary

- Expand UK `ProgramStatistics` coverage for detailed-budget parity by adding `fuel_duty`, `state_pension`, `ni_employer`, and `tax_credits`.
- Add those UK variables to the default model output materialization list so analysis outputs can calculate them without caller-supplied `extra_variables`.
- Validate UK program-statistics configuration before running analysis, with clearer errors for missing model variables or missing materialized outputs.
- Add focused UK program-statistics tests, including a public `economic_impact_analysis()` wiring test with non-program outputs stubbed.

## Notes

- `tax_credits` intentionally overlaps with `working_tax_credit` and `child_tax_credit`; downstream budget adapters should choose the rows they need rather than summing all UK program-statistics rows.
- This PR does not add a full UK integration simulation run. The tests use mocked output datasets to keep PR runtime bounded.

## Tests

- `uv run --extra dev python -m pytest tests/test_uk_program_statistics.py tests/test_us_program_statistics.py -q`
- `uv run ruff check tests/test_uk_program_statistics.py src/policyengine/tax_benefit_models/uk/analysis.py src/policyengine/tax_benefit_models/uk/model.py`
- `uv run ruff format --check tests/test_uk_program_statistics.py src/policyengine/tax_benefit_models/uk/analysis.py src/policyengine/tax_benefit_models/uk/model.py`


anth-volk:
@vahid-ahmadi I requested your review here. Could you specifically check whether the UK programs added in this PR align with the programs currently in use in the UK model/reporting flow, especially the `tax_credits` row and its relationship to `working_tax_credit` and `child_tax_credit`? The PR currently documents that these rows overlap and expects downstream adapters to choose the row set they need rather than summing all of them.

PavelMakarchuk (approved):
Approving. The four new programs (`fuel_duty`, `state_pension`, `ni_employer`, `tax_credits`) bring `economic_impact_analysis(...).program_statistics` to parity with `policyengine-app`'s `UKDetailedPrograms` schema (`src/schemas/societyWideModules.js`). Entities, `is_tax` flags, and entity_variables wiring all line up. `ProgramStatistics.run()` itself is unchanged — this is pure additive wiring plus a pre-flight validation helper.

Test pattern is a nice step up from #358: the mocked simulation builds real `MicroDataFrame` / `UKYearData` / `PolicyEngineUKDataset` objects, with monkeypatching limited to `Simulation.ensure` and the unrelated non-program outputs.

Non-blocking suggestions for follow-up:

1. **Hoist `resolve_entity_variables(simulation)` out of the program loop** in `_validate_program_statistics_config` — it currently runs once per program × 2 simulations and returns the full dict each time.
2. **Tighten the `UK_PROGRAMS` type hint** — `dict[str, dict[str, bool]]` or a `TypedDict` would self-document.
3. **`tax_credits` overlap with `working_tax_credit` + `child_tax_credit`** — matches the legacy app behavior (the app only consumes `tax_credits` from `UKDetailedPrograms`), but worth either dropping the disaggregated rows or marking them with an `overlapping=True` flag in the data structure rather than relying on the PR description comment alone. Today a downstream consumer that naively sums all `program_statistics` rows will double-count.
4. **`_format_missing_program_variables` is a one-liner used once** — could be inlined.
5. **No example update** in `examples/` — small miss.
6. **Schema asymmetry** — `wealth_decile_impacts`/program-stats list remains UK-only on `PolicyReformAnalysis`. Same follow-up as flagged on #358.

PR #360, opened 2026-05-14
Add labor supply response macro output

### LINKED ISSUE #355: Add 4.x output support for legacy labor_supply_response macro result
## Context

`policyengine-api-v2` needs a 4.x-supported way to produce the legacy single-year macro `labor_supply_response` section. Pre-1 `Simulation.calculate_economy_comparison()` returned this in the public macro output. In `policyengine` 4.x, country packages still contain labor-supply-response variables/parameters, but I did not find a first-class output helper that reconstructs the legacy response object.

This blocks the simulation API from migrating back to 4.x without changing the public `SingleYearMacroOutput` contract used by API/report callers.

## Proposed work

Add a 4.x output helper that computes the legacy `labor_supply_response` result, including the aggregate fields and decile/hour breakdowns expected by existing callers.

## Acceptance criteria

- Provide a `.py` output helper for `labor_supply_response` or a documented replacement.
- Match the pre-1 output shape and sign conventions used by `calculate_economy_comparison()`.
- Cover US and UK behavior, including cases where response parameters are disabled or zero.
- Add tests for aggregate totals, decile breakdowns, and hours fields.

anth-volk:
Fixes #355

## Summary

- Add a legacy-compatible `LaborSupplyResponse` output model and calculator for US and UK macro analyses.
- Wire `labor_supply_response` into US and UK `PolicyReformAnalysis` outputs.
- Materialize active LSR-only variables only when policy or dynamic inputs indicate an LSR run.
- Preserve inactive/zero-output behavior without requiring optional LSR columns.
- Harden parameter activation and household-weighted decile calculations against false-positive parameter prefixes and non-default household indexes.

## Tests

- `ruff check src/policyengine/outputs/labor_supply_response.py tests/test_labor_supply_response.py`
- `uv run --extra dev pytest -q tests/test_labor_supply_response.py`
- `git diff --check`

### DIFF
--- a/changelog.d/add-labor-supply-response-output.added.md
+++ b/changelog.d/add-labor-supply-response-output.added.md
@@ -1,0 +1,1 @@
+- Add a legacy-compatible labor-supply response output for US and UK macro analyses.
--- a/src/policyengine/core/dynamic.py
+++ b/src/policyengine/core/dynamic.py
@@ -14,6 +14,14 @@ class Dynamic(BaseModel):
     description: Optional[str] = None
     parameter_values: list[ParameterValue] = []
     simulation_modifier: Optional[Callable] = None
+    affects_labor_supply_response: Optional[bool] = Field(
+        default=None,
+        description=(
+            "Whether this dynamic should materialize labor-supply response "
+            "outputs. None preserves conservative detection for unmarked "
+            "simulation modifiers."
+        ),
+    )
     created_at: datetime = Field(default_factory=datetime.now)
     updated_at: datetime = Field(default_factory=datetime.now)
 
@@ -39,9 +47,22 @@ class Dynamic(BaseModel):
         elif other.simulation_modifier is not None:
             combined_modifier = other.simulation_modifier
 
+        affects_labor_supply_response = None
+        if (
+            self.affects_labor_supply_response is True
+            or other.affects_labor_supply_response is True
+        ):
+            affects_labor_supply_response = True
+        elif (
+            self.affects_labor_supply_response is False
+            and other.affects_labor_supply_response is False
+        ):
+            affects_labor_supply_response = False
+
         return Dynamic(
             name=f"{self.name} + {other.name}",
             description=f"Combined dynamic: {self.name} and {other.name}",
             parameter_values=self.parameter_values + other.parameter_values,
             simulation_modifier=combined_modifier,
+            affects_labor_supply_response=affects_labor_supply_response,
         )
--- a/src/policyengine/core/policy.py
+++ b/src/policyengine/core/policy.py
@@ -14,6 +14,14 @@ class Policy(BaseModel):
     description: Optional[str] = None
     parameter_values: list[ParameterValue] = []
     simulation_modifier: Optional[Callable] = None
+    affects_labor_supply_response: Optional[bool] = Field(
+        default=None,
+        description=(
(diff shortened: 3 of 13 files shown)


anth-volk:
@vahid-ahmadi, when you review this PR, could you be particularly concerned with how `labor_supply_response` impacts integrate into the broader macro output set?

The key risk is that LSRs depend on a conditionally present set of variables. When a reform/dynamic activates labor-supply responses, all of those variables need to be present in the output data for LSRs to calculate correctly; when LSRs are inactive, we intentionally avoid requiring those optional columns and return the legacy zero-shaped result.

I did not add a full integration test for this wiring because the full simulation runtime is long, so the current coverage is focused unit coverage around activation detection, missing-column behavior, output shape, and household-weighted decile calculations.

PavelMakarchuk (approved):
Approving on methodology. Verified field-by-field against legacy `policyengine-api/policyengine_api/endpoints/economy/compare.py:26-98`:

| Field | Match |
|---|---|
| `substitution_lsr` / `income_lsr` (scalar diffs) | ✓ |
| `relative_lsr` = `lsr_hh.sum() / original_earnings.sum()` | ✓ (PR uses `_safe_ratio` → 0 instead of NaN/inf on zero denominator — minor robustness improvement) |
| `decile.average.{income,substitution}` — `.groupby(decile).mean()`, all deciles | ✓ |
| `decile.relative.{income,substitution}` — sum/sum per decile, filtered `> 0` | ✓ |
| `original_earnings = baseline_earnings - total_lsr_hh` (where `total_lsr_hh = reform_lsr - baseline_lsr`) | ✓ (preserves the quirky legacy formula) |
| `hours.{change, income_effect, substitution_effect}` | ✓ |
| `revenue_change = 0.0` | ✓ (preserves legacy bug — `budgetary_impact_lsr` was never populated) |

The `test_calculate_us_labor_supply_response_uses_legacy_shape` test pins hand-computed expected values (e.g. `relative_lsr.income = 40.0/285.0`, `decile.relative.income[1] = 20.0/165.0`), which locks in the legacy numerator/denominator structure.

The activeness-gating piece is new — it skips materializing 6 (US) or 2 (UK) expensive person-level variables when neither policy nor dynamic looks LSR-related. Detection uses parameter-prefix matching (both `labor`/`labour` spellings for UK), an explicit `affects_labor_supply_response` field, and `simulation_modifier` presence as a conservative fallback. Sensible heuristic.

Non-blocking follow-ups:

1. **`revenue_change` legacy bug** — please file an issue to actually compute the LSR revenue effect. Parity with the legacy zero is fine today, but it's a real user-visible gap.
2. **Code volume** — 633 lines for ~50 lines of legacy logic. The active/inactive bifurcation duplicates much of the calc; could be unified by constructing zero-filled series upfront and running one code path. Tests pin behavior so refactoring later is safe.
3. **`_calculate_hours` control flow** — the dual-path column-probing plus `active` flag is convoluted. Likely simplifies to "if active, compute; else zeros."
4. **`affects_labor_supply_response: Optional[bool]` on Policy/Dynamic** — new public API surface for a single use case, with a non-trivial three-way truth table in `__add__`. Consider whether this belongs in a metadata dict instead of a top-level field.
5. **`_safe_ratio` vs legacy NaN/inf behavior** — deliberate improvement; worth a code comment noting this as a deliberate divergence from "legacy-compatible."

PR #322, opened 2026-04-27
Expose axes in household calculators

MaxGhenis:
## What changed

- Adds an `axes=` keyword to `pe.us.calculate_household` and `pe.uk.calculate_household`.
- Accepts either the lower-level nested axes shape or a flat list of axis dictionaries.
- Defaults missing axis `period` values to the household calculator `year`.
- Returns axis-series lists instead of scalars when axes are present, while preserving per-person result objects.
- Documents the new household axes API and adds US/UK regression coverage.

## Why

GiveCalc needs to evaluate the same household across donation amounts. Exposing axes through the public `policyengine.py` household calculators avoids forcing applications to reach into country-package `Simulation` classes directly.

## Checks

- `uv run ruff check src/policyengine/tax_benefit_models/common src/policyengine/tax_benefit_models/us/household.py src/policyengine/tax_benefit_models/uk/household.py tests/test_household_impact.py`
- `uv run ruff format --check src/policyengine/tax_benefit_models/common src/policyengine/tax_benefit_models/us/household.py src/policyengine/tax_benefit_models/uk/household.py tests/test_household_impact.py`
- `uv run --extra dev pytest tests/test_household_impact.py -q`
- `uv run --extra dev pytest tests/test_household_calculator_snapshot.py -q`

PR #363, opened 2026-05-16
Expose axes in household calculators

MaxGhenis:
## Summary

- add `axes` support to the US and UK `calculate_household` helpers
- normalize flat or nested axis specs and default missing periods to the household calculation year
- return per-axis series for person and group entity results

## Why

The app and Python household calculators need the same axis-grid behavior for simple one-household charts. The previous draft PR for this work, #322, is conflict-stuck against current main; this branch refreshes the same useful change onto the current package surface.

## Validation

- `bash .github/check-changelog.sh`
- `ruff format --check .`
- `ruff check .`
- `uv run --extra dev python -m pytest tests/test_household_impact.py -q`

### DIFF
--- a/changelog.d/household-axes.added.md
+++ b/changelog.d/household-axes.added.md
@@ -1,0 +1,1 @@
+Expose axes in the US and UK household calculator helpers.
--- a/docs/households.md
+++ b/docs/households.md
@@ -103,6 +103,44 @@ result = pe.us.calculate_household(
 )
 ```
 
+## Axes
+
+Use `axes` to evaluate one household across a grid of input values. Pass either
+the lower-level nested shape or a flat list of axis dictionaries; missing
+`period` values default to `year`.
+
+```python
+result = pe.us.calculate_household(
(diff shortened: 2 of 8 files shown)


MaxGhenis:
Disposition after full merits review: rebase-and-revive — done. The branch was MERGEABLE and its own 90 lines of axes tests all passed in CI; the only red was tests/test_models.py's label assertion, which main fixed with a tolerant regex (ab381161) about four hours after this branch was cut. Rebased onto current main (no conflicts — the patch-target files were untouched since May) and hardened the one future-brittle assertion (exact income-tax liabilities → the donations-reduce-tax relationship). 68/68 tests pass locally against policyengine-us 1.752.2, including empirical confirmation that every model-version attribute normalize_axes reads exists on the current pin. Design reviewed against the repository public-surface rules: rides the wrapper Simulation (no core bypass), JSON-serialisable list values, kwarg allowlist + error message updated in lockstep. CI re-running now.

PR #365, opened 2026-05-17
Add policyengine.derivations for per-variable computation explanations

MaxGhenis:
## Summary
Adds a new `policyengine.derivations` module so any caller can ask "how did PolicyEngine arrive at this value?" for a single household and get either a structured object or a plain-prose walkthrough.

```python
from policyengine.derivations import derive, narrate

derivation = derive(simulation, "income_tax_before_credits", 2026)
print(derivation.value)                       # 3220.0
print(derivation.trace_text(max_depth=4))     # indented dependency tree
print(derivation.top_level_contributions())   # [(name, value), ...]
narrative = narrate(derivation, country="us", household_summary="TX, joint")
```

## What's in it
- `Derivation` and `TraceNode` — frozen dataclasses for the pruned computation tree (deterministic).
- `derive(simulation, variable, period)` — runs the calculation with `simulation.trace = True`, captures the tree, returns a `Derivation`.
- `top_level_contributions(derivation)` — the immediate dependencies of the root, useful for "breakdown by component" tables.
- `narrate(derivation, ...)` / `narrate_async(...)` — optional LLM narration via LiteLLM. Lazy-imports `litellm` so the deterministic path has no LLM dependency.

## Why this is upstream
[policybench](https://github.com/PolicyEngine/policybench) needed per-cell "PolicyEngine derivation" panels in its prediction-detail modal. The deterministic primitives are general — calculators, dashboards, and papers can use them too — so they belong in `policyengine.py` rather than being baked into a downstream consumer. policybench will be refactored to consume this module.

## Test plan
- [x] Hermetic tests for `is_zero_value`, zero-subtree pruning, and `narrate`/`narrate_async` (with a fake LiteLLM client). The PolicyEngine-US integration tests in `tests/test_derivations.py` exercise `derive` against a real one-adult household.
- [x] `ruff check` + `ruff format` clean.

### DIFF
--- a/changelog.d/derivations-narrative.added.md
+++ b/changelog.d/derivations-narrative.added.md
@@ -1,0 +1,1 @@
+Add ``policyengine.derivations`` for per-variable computation explanations: ``derive(simulation, variable, period)`` returns a structured ``Derivation`` (with pruned trace and top-level contributions); ``narrate(derivation)`` optionally hands it to an LLM for a plain-prose walkthrough.
--- a/src/policyengine/derivations/__init__.py
+++ b/src/policyengine/derivations/__init__.py
@@ -1,0 +1,35 @@
+"""Derivations: structured + narrated explanations of one variable's value.
+
+A ``Derivation`` is the pruned, deterministic computation tree for a single
+``(simulation, variable)`` pair. The tree is the same information OpenFisca
+already records when ``simulation.trace`` is on, but presented as a stable
+data class (independent of OpenFisca internals) so callers can:
+
+- print or serialize the structured tree (deterministic, free),
+- pull out top-level contributions for charts or tables, and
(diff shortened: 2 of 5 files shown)


MaxGhenis:
CI failure (Test 3.13/3.14) is a pre-existing upstream issue: `policyengine_core`'s new `check_computation_modes` rejects `policyengine_us.self_employment_income` because it uses both `adds` and `uprating`. The same failure reproduces on a stale main checkout because the test workflow installs the latest published `policyengine` (which transitively pulls the broken combo), and it is independent of this branch's changes — the new code only adds the `derivations` module.

The hermetic unit tests for `derivations` pass; the integration test that exercises `derive` on a real US Simulation also passes locally with a working `policyengine_us` install.

PR #482, opened 2026-07-22
Sync uv.lock with the 4.22.0 pyproject version

MaxGhenis:
Main's `uv.lock` pins `policyengine 4.20.1` while `pyproject.toml` says 4.22.0 — every `uv run` regenerates the one-line delta and dirties the tree (it kept sneaking into unrelated diffs during the Build N certification; deliberately excluded there). One-line lock sync, no dependency changes.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -2820,7 +2820,7 @@ wheels = [
 
 [[package]]
 name = "policyengine"
-version = "4.20.1"
+version = "4.22.0"
 source = { editable = "." }
 dependencies = [
     { name = "diskcache" },


MaxGhenis:
No checks appear because `pr_code_changes.yaml`'s paths filter (src/tests/scripts/.github/changelog.d/pyproject/manifest) doesn't include `uv.lock` — a lock-only diff triggers no CI by design. Full diff verified: exactly one line, `policyengine` version `4.20.1` → `4.22.0`, matching pyproject. (Side observation for a future touch: the filter gap also means CI never validates lock consistency.)

